PR #1, opened 2023-06-27
add write to sd class

sneakers-the-rat:
I modificied hundreds of files and buried in there somewhere is a treat!


sneakers-the-rat:
(this was a joke to demonstrate how tests help avoid accidental reversions)

PR #2, opened 2023-07-14
v0.1.1 - frame_count

sneakers-the-rat:
Mostly an example of a pull request with tests and docs update

# 0.1.1 - 23-07-13

## Additions

- Added {class}`~miniscope_io.exceptions.EndOfRecordingException` when attempting to read past last frame
- Added {attr}`~miniscope_io.io.SDCard.frame_count` property inferred from the number of buffers and buffers per frame
- Return `self` when entering {class}`~.SDCard` context
- Optionally return {class}`~miniscope_io.sd.DataHeader`s from frame when reading

## Bugfixes

- Index the position of the 0th frame in {attr}`~.SDCard.positions`
- reset internal frame counter to 0 when exiting context


t-sasatani:
Oh, I meant to approve the commits from @sneakers-the-rat and accidentally approved mine too but I suppose it's ok.

sneakers-the-rat:
haha super normal. yeah this is one reason to use feature branches and then pull to dev - your commits to dev that came after mine updated the pull request.

Issue #6, opened 2023-08-13
SD card drive/image selection


t-sasatani:
Miniscope v4 Wire-Free SD cards don't have a file system and seem to have trouble with the Path commands.
Currently avoiding issues by inputting the device path and adding a "directpath" flag in the SDCard constructor, but I feel there might be a better way.
https://github.com/Aharoni-Lab/miniscope-io/commit/bc9370a612ae69b9d182f08f8379762026004ba6

PR #8, opened 2023-09-02
Feature - Battery Log, Data Formats, Plots

sneakers-the-rat:
New features: 
- Data models! Hold a collection of frames and get their headers!
- Plots! Mimic the plots from ye olde notebook!
- Update to pydantic
- Version field in formats
- Format for miniscope firmware with battery voltage level

Reverted:
- `grab_frames` notebook is restored to using the example data and having the evaluated output present


t-sasatani:
> grab_frames notebook is restored to using the example data and having the evaluated output present

The common usage is directly reading SD cards so I think that should be available in the notebook. Although I understand it's better to have a running example with data in the repo is better. So I think we should leave an exposed option for selecting drives.

sneakers-the-rat:
The notebook is just an example, so someone using the code could replace the path which could be a path to a drive, we could just add a comment like you had explaining that explicitly in case it's unclear

sneakers-the-rat:
if we wanted to make the notebooks more than examples, like the intended way to use the package, then I have made a `widgets` module that we could add a file selection dropdown for, and also take the example in `read_frames` which makes an ipywidgets video player as well and make it reusable, but ya probably shouldn't hardcode local variables in the repo

t-sasatani:
That makes sense, but there was a minor problem resolving the device path (https://github.com/Aharoni-Lab/miniscope-io/issues/6), which requires a tiny modification in the later function call. Most initial use cases will directly read SD cards, so it makes sense to have a bit of documentation on that.

sneakers-the-rat:
huh, don't know why that would be the case, `resolve` should only change the path if it's not an absolute path, but we can just take out the `resolve` call, usually it's neutral and just to make problems from using relative paths explicit, but sure we can take that out

t-sasatani:
That makes sense. Would you like to include that commit in this PR or another PR?

sneakers-the-rat:
Here, i'll merge this then make that its own PR, 1 sec

PR #9, opened 2023-09-04
bugfix: relative and absolute paths

sneakers-the-rat:
Previously, sdcard would try and `resolve` every path given to it. Now that behavior has been removed, allowing both relative and absolute paths.

reinstated throwing an exception when config can't be read, as that should stop any further use of the object since something has gone wrong with reading from it

Related to: 
- https://github.com/Aharoni-Lab/miniscope-io/pull/8#issuecomment-1704502701
Closes:
- https://github.com/Aharoni-Lab/miniscope-io/issues/6

@t-sasatani I removed the try block around reading the config which was also added in https://github.com/Aharoni-Lab/miniscope-io/commit/bc9370a612ae69b9d182f08f8379762026004ba6 - i figured removing the `resolve` addresses the reason it was added? seems like throwing an I/O error there if the config can't be read is good to do, since the object shouldn't be used after that point if the file can't be read.

### DIFF
--- a/miniscope_io/io.py
+++ b/miniscope_io/io.py
@@ -32,7 +32,7 @@ class SDCard:
             layout: SDLayout
         ):
 
-        self.drive = Path(drive)
+        self.drive = drive
         self.layout = layout
 
         # Private attributes used when the file reading context is entered
--- a/miniscope_io/io.py
+++ b/miniscope_io/io.py
@@ -29,15 +29,10 @@ class SDCard:
     def __init__(
             self,
             drive: Union[str, Path],
-            layout: SDLayout,
-            directpath: bool = False
+            layout: SDLayout
         ):
 
-        if directpath == False:
-            self.drive = Path(drive).resolve()
-        else:
-            self.drive = drive
-    # Logs the error appropriately. 
+        self.drive = Path(drive)
         self.layout = layout
 
         # Private attributes used when the file reading context is entered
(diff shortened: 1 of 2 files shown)


t-sasatani:
I'm unsure why and am still investigating, but a permission error occurs at the SDcard constructor when using an external drive.

1. I can't locate something that could cause this in the code, but I'm assuming it's not a problem in my environment because of (2).
2. The file can be opened repeatedly outside of the constructor. This is an example notebook that shows this (https://github.com/Aharoni-Lab/miniscope-io/blob/bigfix-permissionerror-example/notebooks/grab_frames.ipynb)
3. This has yet to happen for v0.1.3, though I've tried both of them multiple times.

t-sasatani:
> seems like throwing an I/O error there if the config can't be read is good to do, since the object shouldn't be used after that point if the file can't be read.

Forgot to answer this. I just put it there because io errors happen a lot here (selecting the wrong drive number, admin privileges, driver issues, etc..)

t-sasatani:
I'm sure there are better solutions than this, but changing,
```
        self.drive = Path(drive)
```
to
```
        self.drive = drive
```
makes it work both for the external drive and the image file (https://github.com/Aharoni-Lab/miniscope-io/pull/9/commits/6b929b1e1c0111c89eade2a0e55daca01c9cc7e1). I believe it's because the SD card's path can not be resolved properly and this could be a tentative solution.

PR #10, opened 2023-09-04
Feature - Add video export

sneakers-the-rat:
- Added `to_video` method to SDCard to export video using opencv (already a dependency)
- Moved `skip` to general methods block (no change)
- Added notebook demonstrating save_video
- Added `utils` module with function for hashing files
- Added tests for video export


sneakers-the-rat:
Also in my trying to get tests to work i added testing for macos and windows

t-sasatani:
confirmed it works with external drive (SD card)

PR #12, opened 2023-09-18
New feature - to_img, makes truncated .img files :)

sneakers-the-rat:
Also:
Features:
- Added ReadHeaderException, raised when header cannot be read with a little more information than the IndexError raised currently
- Uses ReadHeaderException to also better detect another kind of currently unhandled end of recording exception
- Automatically switch to using correct file extension in to_video
- Additional `wirefree_battery_sample.img` file to test different formats

Bugfixes:
- When returning with headers, data was incorrectly assigned to "frame" rather than "data", dropping it. Fixed.


miniscope_io/models/data.py
```diff
@@ -18,7 +18,7 @@ class Frame(BaseModel, arbitrary_types_allowed=True):
     Typically returned from :meth:`.SDCard.read`
     """
 
-    data: np.ndarray
+    frame: np.ndarray
```
sneakers-the-rat, line 21: This is the only breaking change - and this is only used in one place internally

sneakers-the-rat:
I'm gonna bypass the review requirement here just bc this is so old and we should get it over with - doesn't affect any shared code

t-sasatani:
My bad I completely forgot about this and the other PR.

PR #13, opened 2023-09-18
Docs wirefree plots complete

sneakers-the-rat:
Added a rendered notebook demonstrating plotting headers!

Features:
- added a "size" arg to plot_headers to explicitly set size to make it match existing notebook

Bugfixes:
- Individual plots don't get multi-axis containers, so don't try and index them like `ax[0]`

Depends on:
- https://github.com/Aharoni-Lab/miniscope-io/pull/12


sneakers-the-rat:
same thing here, just going to merge this because it's very old and also just a docs PR

PR #14, opened 2023-10-23
feat: add support for wireless fpga-based manchester daq

phildong:
add support for wireless fpga-based manchester daq

TODO before merging:
- [x] Test with Opal kelly FPGA Board
- [ ] Test with UART interface


phildong:
as the commit message suggests, still need to figure out how to specify dependencies and include opal kelly libraries

sneakers-the-rat:
Alright! lets start with docs - can we get some module, class, and method-level docstrings to let us know what this module, class, and methods do? The docs are set up to use [google style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) but regular style should work too

sneakers-the-rat:
Added dependency for `bitstring` , i see that `routine` is the module that has the opal kelley stuff in it ( https://github.com/phildong/wireless_fpga_daq/tree/decoder/python/source/routine ) - where did that come from? what kind of license does it have? is it something we could package here, or would we need to refer people to download it elsewhere?

sneakers-the-rat:
Enabled building docs for this branch so you can see the state of the docs as you write them:
https://miniscope-io.readthedocs.io/en/wireless_fpga/
https://miniscope-io.readthedocs.io/en/wireless_fpga/api/uart_daq.html

see build status here:
https://readthedocs.org/projects/miniscope-io/builds/

sneakers-the-rat (changes requested):
Nice - so the main things that are must fix
* sourcing the opal kelley module
* sourcing the bitfile

Then there are things that are like highly recommended we should fix this sooner rather than later, low effort high reward things:
* Docs!!!!! Let's make a habit of documenting things as we write them. Let's not get into a place where when it comes time to release this we have a shitload of docs debt and need to rush through that. let's write them as we go so that we can communicate what things are doing to ourselves/avoid needing to read the source to see what something is doing.
* Unify logging as a function in another module - presumably other classes/modules will want logging so we might as well set that up now, since we init loggers in like 6 different places in different ways here.
* Make dataclasses for configs - move them out of hardcoded variables and function signatures into separate classes so that we have a clear division between configuration, state, and runtime code. 
* Split out opencv plotting from reading logic

Then lower priority, could happen now or later:
* Let's think about what we want as far as making a uniform API across these different I/O modules - eg `SDCard` uses `read() -> np.ndarray` rather than `capture() -> None:` As we start splitting up different modules, that will be necessary to have a clear structure there so we don't need to special case a bunch of things in eg. the opencv reader methods `if hasattr(cls, capture):... elif hasattr(cls, read):...`
* refactor CLI argparsers and add them to the `scripts` in  `pyproject.toml` - maybe move all CLI entrypoints to their own module so that we can keep track of them.
* Tests obvi, but not sure on if this is a "lets merge this ASAP" kind of PR, but ya in the same way as docs, should avoid accumulating technical debt when possible.

sneakers-the-rat:
To make this testable, we would want to split out the operations that are happening in the individual functions from their iteration, which would be good to do generally - so rather than having a method that does some setup and then terminates in a `while 1: ...` loop, we have the method take arguments and return the thing that would go in the queue, and then have maybe even a single function or class that can be used to loop each of those methods getting and putting in the receive/send queues. 

As is these methods would be sort of hard to test because of the way they are dependent on a ton of setup state, but it woudln't be *that* hard if eg. the queues or the multiprocessing `Manager` were stashed as class attributes that we could grab in a testing context.

phildong:
ok I can answer the questions as much as I can, but in general, I'm wondering if we should do all the improvements in this PR or if we should separate them out?
Ideally this PR is about adding fpga support rather than improving the script in general. And this PR already contains a lot more than there should be (ideally I should just plug in the fpga receiver code but that didn't work out).
I'm happy to do everything in this PR and change the name to something like DAQ 2.0 lol but that will take much longer and I'm not sure if that fits @t-sasatani 's timeline.

phildong:
> Added dependency for `bitstring` , i see that `routine` is the module that has the opal kelley stuff in it ( https://github.com/phildong/wireless_fpga_daq/tree/decoder/python/source/routine ) - where did that come from? what kind of license does it have? is it something we could package here, or would we need to refer people to download it elsewhere?

it's proprietary and sort of behind a paywall (you can download for free only if you register and prove you've purchased opal kelly hardware), and there is no easy installer (some manual copy-pasting needs to happen after you download a zip from opalkelly), this is why initially I just git tracked the proprietary codes, but it's probably no longer a good idea.

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+
 
 class uart_daq:
-    def __init__(self, frame_width: int = 304, frame_height: int = 304, preamble = b'\x78\x56\x34\x12'):
+    def __init__(
+        self,
+        frame_width: int = 304,
+        frame_height: int = 304,
+        preamble=b"\x12\x34\x56",
+        header_fmt={
```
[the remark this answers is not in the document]
phildong: will add in docs later but these tuples are basically start and end bit position of a number in header. Not sure if that fits the definition of Range? (at least it should be start/end rather than min/max?)

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+    ):
         self.frame_width = frame_width
         self.frame_height = frame_height
-        self.preamble = preamble
+        self.preamble = Bits(preamble)
+        self.header_fmt = header_fmt
+        self.header_len = header_len * 32
+        self.LSB = LSB
+        self.buffer_npix = buffer_npix
+        assert frame_height * frame_width == sum(self.buffer_npix)
```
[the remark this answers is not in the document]
phildong: tried to compute as much as possible, we can revisit once docs are in

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+        LSB=True,
+        buffer_npix=[20432, 20432, 20432, 20432, 10688],
+        pix_depth=8,
+    ):
         self.frame_width = frame_width
         self.frame_height = frame_height
-        self.preamble = preamble
+        self.preamble = Bits(preamble)
+        self.header_fmt = header_fmt
+        self.header_len = header_len * 32
```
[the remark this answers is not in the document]
phildong: because each word in header is 32 bits and by tradition we talk about number of words in header instead of number of bits

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+        self.LSB = LSB
+        self.buffer_npix = buffer_npix
+        assert frame_height * frame_width == sum(self.buffer_npix)
+        self.nbuffer_per_fm = len(self.buffer_npix)
+        self.pix_depth = pix_depth
+
+    def _parse_header(self, buffer, truncate=False):
+        pre_len = len(self.preamble)
+        if self.LSB:
+            assert buffer[:pre_len][::-1] == self.preamble
```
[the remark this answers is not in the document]
phildong: will definitely add in a message, but I do want to whole thing to fall apart and stop during runtime if this assertion is False cause it really shouldn't be

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+        else:
+            assert buffer[:pre_len] == self.preamble
+        header_data = dict()
+        for hd, bit_range in self.header_fmt.items():
```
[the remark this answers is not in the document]
phildong: I have no clue how to do that, could you give an example or point me to some documentation?

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+        )
+        file.setLevel(logging.DEBUG)
+        fileformat = logging.Formatter(
+            "%(asctime)s:%(levelname)s:%(message)s", datefmt="%H:%M:%S"
+        )
+        file.setFormatter(fileformat)
+        locallogs.addHandler(file)
+        coloredlogs.install(level=logging.INFO, logger=locallogs)
+        # set up fpga devices
+        dev = okDev()
```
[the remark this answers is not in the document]
phildong: not sure exactly what you mean, `okDev` is already a separate class?

sneakers-the-rat:
> ok I can answer the questions as much as I can, but in general, I'm wondering if we should do all the improvements in this PR or if we should separate them out?

ya i split review comment into "must fix," "should fix," and "nice to have" for this reason - i think resolving the two basic mustfixes should be done before merging (because the `routine` import would break the existing uart code, and without the bitfile the fpga part wouldn't work), and then we can address the others in separate PRs if we need this to be merged ASAP.

> it's proprietary and sort of behind a paywall (you can download for free only if you register and prove you've purchased opal kelly hardware)

Is it distributed with a license? Wondering what we have in mind for this package - it's not unheard of to need to manually download some proprietary plugin, but its definitely not the most convenient thing. At least we need to provide docs about what it is and where to get it. If the license allows us to just package it here that would be ideal.

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
+        time.sleep(0.01)
+        dev.setWire(0x00, 0b0)
+        # read loop
+        cur_buffer = BitArray()
+        pre = Bits(self.preamble)[::-1]
+        while True:
+            buf = dev.readData(read_length)
+            dat = BitArray(buf)
+            cur_buffer = cur_buffer + dat
+            pre_pos = list(cur_buffer.findall(pre))
```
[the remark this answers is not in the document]
phildong: how would you zip through two different copies of the same iterator without first casting it to a list? will it essentially find twice in each iteration if you just `zip(it[:-1], it[1:]`?

miniscope_io/uart_daq.py
```diff
@@ -1,211 +1,351 @@
...
     # COM port should probably be automatically found but not sure yet how to distinguish with other devices.
-    def capture(self, comport:str = 'COM3', baudrate:int = 1200000, mode:str = 'DEBUG'):
-        logdirectories = ['log', 'log/uart_recv', 'log/format_frame', 'log/buffer_to_frame']
+    def capture(
+        self,
+        source: str,
+        comport: str = "COM3",
+        baudrate: int = 1200000,
+        mode: str = "DEBUG",
+        bit_file: str = "/Users/mbrosch/Downloads/USBInterface/USBInterface-6mhz-3v3-INVERSE.bit",
```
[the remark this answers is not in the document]
phildong: that's a last-week excitement mistake as it definitely shouldn't contains Marcel's name. It's supposed to point to a relative path within the repo so it will come with the script, but that's no longer possible. I think this can definitely be in a config file, but we just need to make sure users somehow obtain that file and put it in correct location.

phildong:
found the license! https://opalkelly.com/about-us/frontpanel-license/

sneakers-the-rat:
well that sucks

> Licensee shall not modify, distribute, resell or otherwise transfer Software for any purpose, commercial or non-commercial. Licensee may integrate and sub-license Software for distribution but Software must only be used in conjunction with Opal Kelly devices or devices incorporating Firmware.

So that poses a sort of big problem - we can't really... distribute this thing like that. at all.

sneakers-the-rat:
It looks like OE vendors in the C libraries?
https://github.com/open-ephys-plugins/rhythm-plugins/tree/main/Resources

sneakers-the-rat:
Added in the opalkelly stuff, did some minor refactoring to keep vendor code separated from ours, and also made a `devices` module because it seems like we'll need that eventually anyway - ie. the fpga code and the uart code should eventually be separated and their common parts should use composition or inheritance rather than differently named modules.

But as you can see, the tests fail because the module can't be imported. We could do the import manually in the containing `__init__` file, but when i try and import it directly on my mac I get a platform incompatibility error, which is sort of expected. 

@phildong did you compile these yourself, or did the SDK download come with precompiled binaries? If it came with the source, we can just compile them as part of the packaging process as a C extension. Otherwise, we will need standalone binaries for win/mac as well as linux.

phildong:
ok finally got around to work on this. Thank you so much @sneakers-the-rat for setting this up and sorry again for the lack of response!
Just pushed some commits and I think there are 3 things left to do:

1. obtain the binaries for mac and windows. I can download those files but they're packaged in .dmg and .exe respectively, so we either need to find some windows/mac computers or need to figure out how to "unpack" those files on linux. (@MarcelMB should already have the mac files though)
2. figure out how to symlink based on OS. as you can see I pushed some symlink from the ubuntu-specific folder to the lib folder. We either need to figure out how to do this automatically based on OS (`__init__.py` maybe?), or come up with better solution than symlink.
3. figure out how to set environment variables before python execution as documented [here](https://docs.opalkelly.com/fpsdk/frontpanel-api/programming-languages/): 
Screenshot from 2023-10-29 17-43-10
I was able to do this with a bash script: https://github.com/phildong/wireless_fpga_daq/blob/183f8d4a6c33ec0910a9c4791c65fdd72f31311d/python/start_interface.sh but I feel there must be a better solution

at the moment I have no clue what's the best way to do either of those 3 things and how they would interact with the rest of the repo, so I'd love some help before I come up with some very broken and hacky way to do them :rofl:

sneakers-the-rat:
ok so @phildong can you send me a copy of what the SDK comes to you as, fresh from being downloaded from OK? that should help clarify what we need to do here :)

sneakers-the-rat:
Got mac import working, added a README file https://github.com/Aharoni-Lab/miniscope-io/blob/209b560d8909404bcd80bd4f7e536f6a68fd67b1/miniscope_io/vendor/opalkelly/README.md

sneakers-the-rat:
OK Got this working for linux and mac, marked windows as not implemented.

Added documentation on what was done to make them work here: https://github.com/Aharoni-Lab/miniscope-io/blob/111109baf9d25ff0103f13f615ec898f98c47f45/miniscope_io/vendor/opalkelly/README.md

And also added special installation instructions needed for debian installs (this doesn't seem to be necessary in ubuntu, at least according tot he test runner):
https://github.com/Aharoni-Lab/miniscope-io/blob/111109baf9d25ff0103f13f615ec898f98c47f45/docs/guide/installation.md

which is rendered here: https://miniscope-io.readthedocs.io/en/wireless_fpga/guide/installation.html

It would be **great** if y'all could also contribute some docs about how these things work and what they require to work before merging the code, but if this is for some reason a time-critical merge then we can pop another issue to remind us to it later, but we shouldn't leave new additions to the package undocumented or else the thing starts to turn into a pile of spaghetti where we lose visibility into what's actually in it

I have written a *very weak* test function that just tests the import of the module - to make sure this actually works, i would **strongly recommend** that the folks who know how this works add some test code that actually test the functionality of the packaged drivers. This can even be done by marking them as skip when running the tests normally, but runs them only when there actually is an FPGA board on the testing machine. ( see https://docs.pytest.org/en/6.2.x/skipping.html ) 

Otherwise **the only guarantee we have that this code works is vibes** and that is **not a guarantee suitable for code that we distribute.**

sneakers-the-rat:
I think the only thing remaining here thats a must fix is adding the bitfile? @MarcelMB said @phildong generated it? Is that something that is a static file thats always the same, or is it something that might evolve with the FPGA (and thus would need to be versioned?)

sneakers-the-rat:
OK Removed the bitfile from the method args and added some TODO items for documentation.

Also replaced the [mutable defaults](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) in the function signatures with data models that match the models used in the SDCard model. Note for myself in there to think about how that should be structured -- want to get it so all the I/O classes speak the same language. 

I think those are all the must-fixes? so once Takuya approves then i think we can merge?

miniscope_io/uart_daq.py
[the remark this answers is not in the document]
sneakers-the-rat: probs have word size as its own parameter :)
https://github.com/Aharoni-Lab/miniscope-io/blob/88d9e7c6b6b2ae5ccbd92f53e337a2fad3e279fa/miniscope_io/sdcard.py#L114
will make note to self for later PRs

miniscope_io/uart_daq.py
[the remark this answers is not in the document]
sneakers-the-rat: added example: https://github.com/Aharoni-Lab/miniscope-io/pull/14/commits/60ecb3d34d4498137622081eb2a3ad2f4a8f2393

miniscope_io/uart_daq.py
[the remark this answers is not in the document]
sneakers-the-rat: sry i mean the config that follows, the wire settings and etc.

miniscope_io/uart_daq.py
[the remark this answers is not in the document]
sneakers-the-rat: it's probably fine, sorry i was being picky

sneakers-the-rat:
Wat r we waiting on here @phildong @t-sasatani - lets merge this pup so we can get everyone running the same code, I think @MarcelMB is still using the old repo code

miniscope_io/stream_daq.py
```diff
@@ -0,0 +1,804 @@
...
+            Header format used to parse information from buffer header, by default `MetadataHeaderFormat()`.
+        header_len : int, optional
+            Length of header in (32-bit) words, by default 11.
+            This is useful when not all the variable/words in the header are defined in :class:`.MetadataHeaderFormat`.
+            The user is responsible to ensure that `header_len * 32` is larger than the largest bit position defined in :class:`.MetadataHeaderFormat` otherwise unexpected behavior might occur.
+        LSB : bool, optional
+            Whether the sourse is in "LSB" mode or not, by default True.
+            If `not LSB`, then the incoming bitstream is expected to be in Most Significant Bit first mode and data are transmitted in normal order.
+            If `LSB`, then the incoming bitstream is in the format that each 32-bit words are bit-wise reversed on its own.
+            Furthermore, the order of 32-bit words in the pixel data within the buffer is reversed (but the order of words in the header is preserved).
```
phildong, line 122: This is, as far as I understand, what d168e696fccf7f2209d7b9649c74d8cf44f4a0ba is doing, but it really makes no sense to me. @MarcelMB @sneakers-the-rat maybe double-check this is indeed working on the microcontroller with correct firmware?

sneakers-the-rat: pretty much 🤷‍♂️. I think that in future refactors we'll want to be able to coordinate with the firmware/FPGA code and resolve that bit ordering problem so that we don't need this flag, but for now this is fine

t-sasatani: I think I did funky bit order in firmware because:
- It's easier to handle data as 32-bit words in firmware, considering it's a 32-bit MCU.
- Some clock timing issues made receiving 32-bit asynchronous serial (UART) packets unstable, leading me to read out this data as 8-bit words on the DAQ side.

Technically, Manchester encoding should handle this problem better, and since we won't really need UART once we all get Opal Kelly boards, it might make sense to deprecate the UART function for now. This would make merging the PR easier and make it simple for me to fix the firmware in a way that makes more sense.

phildong: sorry I guess github won't allow me to point to a single line and it's leading to some confusion. I figured the 32-bit words are reversed and that has been find for a while.

I was specifically talking about line 122 which seem to be what d168e696fccf7f2209d7b9649c74d8cf44f4a0ba is doing and it's surprising to me, since that implies on top of reversed bit order per word, the words themselves are transmitted in a chronically reversed order (and that's only true for pixels but not header??). Personally I've never seen the wireless data like this when testing with @MarcelMB so just want to make sure this is indeed the case.

sneakers-the-rat: Ya ya thats this:
https://github.com/Aharoni-Lab/miniscope-io/pull/17
https://github.com/Aharoni-Lab/miniscope-io/issues/16

sneakers-the-rat: Looking at that now... should that just be a full array reversal lmao. I think I was just trying to leave as much code in place as possible and not thinking about it

phildong: it should be if, again, we can CONFIRM that bit-wise inversion of the whole buffer (minus the header) IS the intended behavior of MCU.

sneakers-the-rat: fucking lmao ok this is embarassing

```python
import numpy as np

initial_array = (np.random.random((32*32,)) * 256).astype(np.uint8)

# the reverse from the current implementation
weird_reverse = np.array(
    [
        initial_array[i : i + 32][::-1]
        for i in reversed(range(0, len(initial_array), 32))
    ]
).flatten()

# how you might uh normally flip an array
regular_reverse = np.flip(initial_array)

assert np.array_equal(weird_reverse, regular_reverse)
```

sneakers-the-rat: so it is just reversing the whole array. the prior implementation that just reverse the order of the 32-bit chunks while keeping the intra-chunk order was wrong, and all we did in our fix was to reverse the chunks too -- which is equivalent to just flipping the whole array but probably way slower lol

sneakers-the-rat: I'm not sure why we are using the bitstring package there instead of immediately casting to a numpy array, but since the next line is just to read it into a numpy array then i say we just to the flip there lmao. let's not handle that in this PR tho, let's get this one merged and open that as its own issue

phildong: the prior implementation worked for a few instances (including the first wireless transmission video) and is the expected behavior from what I understand.

sneakers-the-rat: I think what we learned was that it appeared to work but we encountered https://github.com/Aharoni-Lab/miniscope-io/issues/16 when trying to stream an actual video (rather than just confirming that there was data being transferred) - @MarcelMB would know the history more, that's about where i came in. IDK if there was some change in the firmware, etc. but this is the kind of thing that is the reason we should have clear API contracts with tests between software/firmware/hardware components :). Go from 'this used to work' to 'we know this works because we have end-to-end tests that confirm that it works'

sneakers-the-rat: anyway, that seems like something for future PRs, returning to this one - there are two items listed in the root comment that are necessary to merge this, testing on the FPGA and testing on the UART, any sense of what it would take to get that done so we can merge this and move onto the next things??

sneakers-the-rat:
so wats da deal with this? there is some interest from @hsemwal in making a python interface to the griniscope, and i'm not sure what @MarcelMB is doing to test the thing, but it would be cool to merge this so we can all be using the same tools together. not sure what the wireless miniscope ppl are thinking re: timeline but it would probably be nice to launch with a polished interface, and so we could do like a joint paper re: redoing the miniscope daq stuff to say 'we not only want to make dope microscopes, but microscopes that are easy to use and maintain.' 

thimkin bout issues like these https://github.com/Aharoni-Lab/Miniscope-DAQ-QT-Software/issues/70 that we can't really deal with bc we don't have a good interface distributed along with the scopes that can do stuff like run diagnostics or ensure system state to avoid common bugs, and the path there goes through having more shared code, and this is an easy, low stakes place to start <3

hsemwal:
I am certainly down to write up an interface for all these Miniscopes. Im thinking something where the program can run different miniscopes (we are generalizing the firmware code to this end), and automatically configuring the program (so no click and dragging Java scripts to change imaging, ie FOV, FPS, gain). maybe we could start this week and have something setup by the start of summer when we are doing in vivo testing?

t-sasatani:
I bought the Opal Kelly board and tried to run this, but found bugs around handing over data between queues. So @MarcelMB should be using the old repo.
- I need to figure out some stuff in the FPGA code before fixing this code.
- I'm thinking of ignoring UART compatibility for now because only I was using it.
- I'm planning to work on these issues soon, but am having trouble finding time to sit on the bench.
- This PR has so many topics mixed up now so I (or maybe @MarcelMB if he's down) will merge this PR as soon as it works on the Opal Kelly board.
- After merging, we'll have a primitive CLI interface and can think about making that better.

sneakers-the-rat:
> Im thinking something where the program can run different miniscopes (we are generalizing the firmware code to this end), and automatically configuring the program (so no click and dragging Java scripts to change imaging, ie FOV, FPS, gain). 

that is what this package is for!

So we would have some base class like

```python
from abc import ABC, abstractmethod


class Miniscope(ABC):

    @abstractmethod
    def set(self, cmd: 'Command') -> 'Result':
        ...

    @abstractmethod
    def get(self, value: 'Value') -> 'Result':
        ...

    @abstractmethod
    def start(self):
        ...

    @abstractmethod
    def stop(self):
        ...

    @abstractmethod
    @property
    def config(self) -> 'Config':
        ...

    @config.setter
    def config(self, key: str, val: 'Value'):
        ...

```

and so on with some minimal set of methods that each miniscope needs to implement, and then use composition to make single command modules for different protocol types, mappings between the common operations for a miniscope and whatever device-specific values are needed to use them, etc. Each of those implied classed like `'Value'` and `'Result'` and etc. can hold enums of possible values, carry results and do stuff like logging and error handling in a sane way, and so on.

Then we could build a GUI on top of that that would be general across scopes (yes, no manually editing and clicking and dragging JSON). you know, fun for the whole family.

So we are already started! merging this PR gets the wireless scope in the repo, and from there we can refine it into a common public API class like that. If we get some sample outputs from the hardware to mock in we coudl ensure that those tests don't break the code as we go, etc. 

Like i'm taking a look at some of the stuff @phildong did here https://github.com/phildong/wireless_fpga_daq/pull/2 and it looks pretty dang good! Not going to link to specific files since it's a private repo, but like all these methods for debugging and analyzing the data could be made into general routines we test and make available to future miniscope users. if we are [actively using the same code we intend to distribute internally as well](https://en.wikipedia.org/wiki/Eating_your_own_dog_food), then it's easier for us in the long term and also makes sure that we don't have any extra secret knowledge that users don't have. otherwise, we get into situations like with the Qt DAQ where since none of us really actively use it and instead have our own personal set of tools to develop with, and it becomes a huge chore on both ends - to debug any problem with the public software, and to maintain 10 different sets of tools. We can make our lives easier <3

---

> I bought the Opal Kelly board and tried to run this, but found bugs around handing over data between queues

thanks for getting to what you could @t-sasatani , i imagine it's hard to find time at the bench getting the new lab started! If @MarcelMB can replicate those then i would be happy to help patch them. 

> I'm thinking of ignoring UART compatibility for now because only I was using it.

Totally cool, we can just mark that as unmaintained and deprecate it after this PR if we end up not needing it, but i think it was still a good exercise in figuring out how to merge overlapping code :)

> This PR has so many topics mixed up now so I (or maybe @MarcelMB if he's down) will merge this PR as soon as it works on the Opal Kelly board.

Totally agreed lol, we should address the final parts of it and then merge it so we can then handle the individual pieces separately in smaller increments. We have the luxury still of this not being used publicly (as far as I know) so we don't have to be so cautious about merging, and can iterate even if it's only partially working

sneakers-the-rat:
Selfishly, I'd like to use this as a practice project to learn Rust: convert the C/C++ in the Qt DAQ repo into rust and then expose it to this package with PyO3, so thats why i am eager to get us working together in this module - i wanna practice writing super fast code to be able to do stuff like running multiple scopes that the other projects will need, but can't do that if all the fun real-world development is happening elsewhere :'(

MarcelMB:
Hey,

I just saw the messages and my mentions here from the last days. For some reason, I didn't get any email notifications. Not sure why.

So I guess the goal of this PR is to see if we can get an actual image transfer working using the Opal Kelly FPGA board and the 'streaming' code here actually gets images. In the wirless_FPGA branch is only Python code
I was just wondering if there is an MCU code here that is also to be used and tested or if I just use the MCU code we have on our Miniscope framework repository and that I used last year fall to do image acquisition?

sneakers-the-rat:
Yes that is the goal of this package - to move away from having hardware, firmware, and software all tightly coupled in the same package to making software that can generalize across firmwares :) use what you intend to release. If it doesnt work, we fix it. By keeping them separate, we enforce a stable API

phildong:
I realized this has been going on too long and I already forgot how things work lol. @sneakers-the-rat could you remind me how did you ended up handling the environment variable setup mentioned here?

> ok finally got around to work on this. Thank you so much @sneakers-the-rat for setting this up and sorry again for the lack of response! Just pushed some commits and I think there are 3 things left to do:
>  ...
> 3. figure out how to set environment variables before python execution as documented [here](https://docs.opalkelly.com/fpsdk/frontpanel-api/programming-languages/):
>    Screenshot from 2023-10-29 17-43-10
>    I was able to do this with a bash script: https://github.com/phildong/wireless_fpga_daq/blob/183f8d4a6c33ec0910a9c4791c65fdd72f31311d/python/start_interface.sh but I feel there must be a better solution

Did it get handled with installation script? If so, what's the best way to set things up for testing of this PR? Something like `pip install +git:xxxxx`?

cc: @MarcelMB

sneakers-the-rat:
We vendored in the opalkelly code to the vendor subpackage, and used it in the devices subpackage.

We want there to be no env variable manipulation needed visible to the user, if possible, hence the vendoring. Patching env variables outside the package is only necessary if the library files can be in a variable location relative to the rest of the code, and we didnt want ppl to have to sign up for an OK account, download it, set env vars, etc. Patching env vars should happen as a part of importing that subpackage, and if we do need to support variable lib locations we should do so in a .env or config file rather than in a bash script. (I get that it was a temporary hack to get things running, not criticizing just stating design goals)

This is a normal python package, and we want it to behave like one - no special launching logic needed, so to run the code yes you can install it from git like that, but you probably should just clone the repo, switch to this branch, install in a venv. So either

```
python -m venv venv
source venv/bin/activate
pip install -e . 
```

or

```
poetry install
poetry shell
```

we can make any sort of entrypoints we want to mimic whatever existing scriptlike behavior yall are using, so like for this PR lets just make sure it works and then we can work on making CLI entrypoints, making windows mode work, etc. In future PRs

MarcelMB:
I am working on it. I was just trying to get data streaming work with the repos we have. Just to check if the hardware is working properly before checking if this PR code is working here. But I can't even connect to the FPGA. There is some problem that I didn't figure out yet. Like I cannot build a connection with the FPGA as I did every other week without changing anything.
So I'm stuck at the step. Find the FPGA at the moment. Which never caused a problem before. 

Screenshot 2024-03-11 at 3 16 14 PM

sneakers-the-rat:
Is that the full traceback? And it looks like thats still the old repo. Part of why we should move to a collectively maintained repo is so we can progressively improve UX problems like these as we work on the same tools we distribute for users, so in this case we would want to a) identify the bug, b) resolve the bug if possible, c) add more informative error message, d) add tests to prevent bug from recurring.

So yes exactly why we should merge this and be able to iterate on problems like these. Moving our level of confidence in the software from "I didnt change anything" to "we have tested code we know works under xyz conditions"

sneakers-the-rat:
Do
```
python -m serial.tools.list_ports
```

To list current ports

MarcelMB:
So this is unfortunately paused because the FPGA is dead. Need to wait until we get the new ones running. Sad but let's hope we can continue very soon

sneakers-the-rat:
RIP FPGA

miniscope_io/stream_daq.py
```diff
@@ -789,13 +802,8 @@ def main():
...
-        try:
-            from miniscope_io.devices.opalkelly import okDev
-
-            HAVE_OK = True
-        except (ImportError, ModuleNotFoundError) as e:
-            warnings.warn(f"Cannot import OpalKelly device, got exception {e}")
+        if not HAVE_OK:
+            raise ImportError('Requested Opal Kelly DAQ, but okDAQ could not be imported, got exception: {ok_error}')
 
         daq_inst.capture(source="fpga")
```
MarcelMB, line 809: pushed HAVE_OK to begining of file and added instance of if not HAVE_OK

MarcelMB:
I set up the hardware for wireless data transfer of test frames today and tested the `stream_daq.py` code

With the great help of @sneakers-the-rat we got the code to run together after multiple errors in the software. Evrything works now with the hardware and actually transmitted data wirelessly. 

Pushed a new commit to this branch. Had to declare some variables earlier in the code. So just moved their position and removed some type annotations. No drastic changes

@phildong @t-sasatani Maybe you can quickly look over the code and let us know if we are good to go to do the Merge with main

[https://github.com/Aharoni-Lab/miniscope-io/pull/14/commits/088c4489d84f9b2e9da5346e0d78e30182264a67](url)

t-sasatani:
Awesome, and thanks! The changes make sense and I think it's ready for merge.

Just one thing: Would it be possible to add an FPGA bitstream file in this repo?
This can be another small PR, but I think you can also throw it in this PR.

sneakers-the-rat:
@MarcelMB if you add the bit file in the ok folder I can make a simple lil provider function

t-sasatani:
Thanks @sneakers-the-rat ! A provider function will be best and it seems like the frontpanel SDK can do that, but if it takes a bit of experimenting on HW, just having a fixed bitstream file here would already be better. The problem on my end is that the synthesis takes near an hour on my mobile PC and it's a bit inconvenient to change parameters and re-upload to find which config works.

sneakers-the-rat:
Totally ya i just meant like a function that returns a path basically, so that theres one layer in between the file and the consuming class so we can eg. Switch between bitfiles later

MarcelMB:
I looked through the stream_daq.py file and couldn't find where it is actually taking the .bit file for the FPGA for?

So I was wondering where the line of code is for taking such a file and uploading it to the FPGA? And how in yesterday's test it even knows at which baud rate we are operating. because we need a different bit file for different speeds of the wireless data transfer. 

I assume this file is uploading the bit files [opalkelly.py](https://github.com/Aharoni-Lab/miniscope-io/blob/wireless_fpga/miniscope_io/devices/opalkelly.py) 

Just asking because right now I don't even know which bit file to upload here since I cant find where it is getting it from.

sneakers-the-rat:
Huh ya theres a method here but its not called in the main script file thing: https://github.com/Aharoni-Lab/miniscope-io/blob/088c4489d84f9b2e9da5346e0d78e30182264a67/miniscope_io/devices/opalkelly.py#L28

We'll want to clean this code up as we go to make a nice and tidy CLI interface, including exposing setting/reading config like bitfiles.

Basically we want to have a versioned collection of all the bitfiles that can/should be used with the device. I still dont know exactly what they are or do, so that should get documented, but ya we want everything that we would expect ppl to use in here.

sneakers-the-rat:
Idk! There is no bitfile upload invoked when running that code, I assume it was running whatever was last on the device. Making bitfile selection part of the startup routine can be a future PR I think. Can we merge this and then iterate on that?

sneakers-the-rat:
https://github.com/Aharoni-Lab/miniscope-io/issues/19

MarcelMB:
for now, I have added a .bit file that was the last one uploaded to the FPGA into the devices folder next to opalkelly.py

I can go ahead and merge the PR. Haven't done that before but I can try

sneakers-the-rat:
hell ya.

t-sasatani:
>I assume it was running whatever was last on the device.
Yup this is correct. It was taking time to generate the same bitstream file on my end.

>I have added a .bit file that was the last one uploaded to the FPGA into the devices folder next to opalkelly.py
So yes this perfect! Thanks!

PR #23, opened 2024-06-14
Linting - add Ruff and Black

sneakers-the-rat:
builds off: https://github.com/Aharoni-Lab/miniscope-io/pull/21
since it's about to be merged.

Seems like we're going to get into trouble with our linting sooner rather than later now that work has begun on this in earnest.

Added black and ruff checks in CI, and started getting the package in line with them. Will add docs on how to do this during development but it's super simple (we can set this up as a pre-commit action, but otherwise it's just `black miniscope_io && ruff check` ).

gotta run for now, but it's super mechanical work, nothing too tricky, and should all be totally function neutral.


sneakers-the-rat:
@t-sasatani any objections to my linter rules here? 

https://github.com/Aharoni-Lab/miniscope-io/blob/beec7d0983200ca96b5189457801e1c00580d717/pyproject.toml#L67-L106

lmk if there are any you find annoying or wish we had

https://docs.astral.sh/ruff/rules/

t-sasatani:
I looked into your code changes (which I believe indicate these rules), and I think they look good. I haven't used these before, so I'm not sure what coding with this will be like, to be honest, but I think we can just start and think later.

sneakers-the-rat:
Ill write some docs on how to set it up to be more or less automatic :) these integrate pretty well with most IDEs

sneakers-the-rat:
I review this pull request and declare it to be good

PR #26, opened 2024-06-18
Tests for streamDaq / video export

t-sasatani:
I drafted a simple test module for the streamDaq because it's starting to take shape. This is pretty preliminary and might be a bit early for reviewing, but I'd appreciate any feedback. Updates are listed in the following:
- Add FPGA RAW data recording and importing mode
- Add video write-out function (to avi) and a handle to turn on and off video saving/display
- Added a primitive Pytest module and confirmed that it works locally (by running normally/intentionally breaking it)

Here is some stuff I'm wondering about and will think about later:
- How to switch between normal and test operations nicely. I guess the current way is fine but the config file and model is getting a bit non-intuitive.
- Asserting metadata. The log output has time in it, so I guess we want some other metadata export, also because we'll need metadata anyway. I can quickly assemble a CSV export for tests, but it would be better to have it compatible with previous Miniscopes if possible.


t-sasatani:
The added pytest module goes through local tests but fails on GH actions, so I'll look into it later.

sneakers-the-rat:
Oh this rocks. Ill do a review in a sec but I definitely know how to do this now that we have sample data and expected values up and running :)

miniscope_io/__init__.py
```diff
@@ -8,6 +8,7 @@
 DATA_DIR = BASE_DIR / 'data'
 CONFIG_DIR = DATA_DIR / 'config'
 DEVICE_DIR = BASE_DIR / 'devices'
+TESTDATA_DIR = BASE_DIR.parent / 'tests' / 'data'
```
sneakers-the-rat, line 11: we'll want this to be in `tests/conftest.py` - this directory won't be present in the wheel or sdist of the package, and is only relevant at test time :)

miniscope_io/models/stream.py
```diff
@@ -16,6 +16,7 @@
 class StreamBufferHeaderFormat(BufferHeaderFormat):
     pixel_count: Range
 
+ALLOWED_MODES: Tuple[str, ...] = ("DAQ", "RAW_RECORD", "RAW_REPLAY")
```
sneakers-the-rat, line 19: this will be a general note i'll make throughout this, but we want to mock the OkDev class rather than modify this one :)

miniscope_io/models/stream.py
```diff
@@ -68,10 +71,18 @@ class StreamDaqConfig(MiniscopeConfig, YAMLMixin):
...
+    save_video_name : str, optional
+        If video is saved, , by default False.
+    show_video : bool, optional
+        Whether the video is showed in "real-time", by default True.
 
     ..todo::
         Takuya - double-check the definitions around blocks and buffers in the firmware and add description.
     """
+
+    mode: Optional[str] = 'DAQ'
```
sneakers-the-rat, line 85: `Optional` is equivalent to `T | None`. since `None` is not a valid value here, this should just be a `str` with a default value.

or, if it is one of the `ALLOWED_MODES` it should be something like `Literal[*ALLOWED_MODES]`

miniscope_io/models/stream.py
```diff
@@ -85,6 +96,22 @@ class StreamDaqConfig(MiniscopeConfig, YAMLMixin):
     block_size: int
     num_buffers: int
     LSB: Optional[bool]
+    save_video: Optional[bool] = False
```
sneakers-the-rat, line 99: same thing here as above, this should just be `save_video: bool = False`

miniscope_io/models/stream.py
```diff
@@ -85,6 +96,22 @@ class StreamDaqConfig(MiniscopeConfig, YAMLMixin):
...
+    test_raw_data_path: Optional[Path] = TESTDATA_DIR / 'stream_daq_test_fpga_raw_input_200px.bin'
+
+    @field_validator('mode', mode='before')
+    def check_mode_string(cls, value: str) -> str:
+        if value not in ALLOWED_MODES:
+            value = 'DAQ'
+        return value
+
+    @field_validator('save_video_filename', mode='before')
+    def check_video_filename_string(cls, value: str, values) -> str:
```
sneakers-the-rat, line 111: this looks like it will cause the model to fail to instantiate with default values, since `save_video = False` and `save_video_filename = output`

miniscope_io/stream_daq.py
```diff
@@ -23,12 +24,11 @@
 ok_error = None
 try:
     from miniscope_io.devices.opalkelly import okDev
-
     HAVE_OK = True
 except (ImportError, ModuleNotFoundError) as ok_error:
     module_logger = init_logger('streamDaq')
     module_logger.warning(
-        "Could not import OpalKelly driver, unable to read from FPGA!")
+        "Could not import OpalKelly driver, you can't read from FPGA!\nCheck out Opal Kelly's website for install info\nhttps://docs.opalkelly.com/fpsdk/getting-started/")
```
sneakers-the-rat, line 31: ideally we have vendored all this so people don't need to install anything extra :)

miniscope_io/stream_daq.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         self.preamble = self.config.preamble
         self._buffer_npix: Optional[List[int]] = None
         self._nbuffer_per_fm: Optional[int] = None
+        self.terminate = multiprocessing.Value('b', False)
```
sneakers-the-rat, line 82: probably use an `Event` or something for this, i think `Value` has a decent amount of overhead (but i could be wrong)

miniscope_io/stream_daq.py
```diff
@@ -215,29 +216,41 @@ def _fpga_recv(
...
+            if self.config.mode == 'RAW_REPLAY':
+                with open(self.config.test_raw_data_path, 'rb') as file:
+                    buf = bytearray(file.read())
+            else:
+                buf = dev.readData(read_length)
+            if self.config.mode == 'RAW_RECORD':
+                fpga_raw_path = 'data/fpga_raw.bin' # Not sure where to define this because we don't want to overwrite.
+                with open(fpga_raw_path, 'wb') as file:
+                    file.write(buf)
+                self.terminate.value = True
```
sneakers-the-rat, line 253: yes here we will want to mock the OK device instead of having an `if else` switch in the method. 

we will also want to make `record`-ing a more generalizable switch rather than a mutually exclusive mode with normal operation, so like we would have a `capture()` method with `capture(record=True)` or `capture(record='binary')` or something

miniscope_io/stream_daq.py
```diff
@@ -247,6 +260,9 @@ def _fpga_recv(
                 serial_buffer_queue.put(cur_buffer[buf_start:buf_stop].tobytes())
             if pre_pos:
                 cur_buffer = cur_buffer[pre_pos[-1] :]
+            if self.config.mode == 'RAW_REPLAY':
+                time.sleep(1) # Non-ideal. Wondering if there's a better way than checking all the queues are empty with a flag or plugging in all queues in this function.
```
sneakers-the-rat, line 264: we want to make this whole chain of queues robust enough that we don't need to sleep here :). for testing purposes, we want to use the exact code that would be used normally and mock the hardware part, bc otherwise we aren't really testing the functionality of the class, but the functionality of the test conditions

miniscope_io/stream_daq.py
```diff
@@ -496,6 +512,17 @@ def capture(
         else:
             raise ValueError(f"source can be one of uart or fpga. Got {source}")
 
+        # Video output settings
+        fourcc = cv2.VideoWriter_fourcc(*'DIVX')
+        frame_rate = 20.0 # this should be in the config yaml
+        frame_size = (self.config.frame_width, self.config.frame_height)
+        out = cv2.VideoWriter('output.avi', fourcc, frame_rate, frame_size)
```
sneakers-the-rat, line 519: see this method here - let's make a generalized `VideoWriter` wrapper class and use it throughout the package for consistent behavior. https://github.com/Aharoni-Lab/miniscope-io/blob/aabd2322af5f60fb91693679a0d4188a6606a771/miniscope_io/io.py#L362

miniscope_io/stream_daq.py
```diff
@@ -496,6 +512,17 @@ def capture(
...
+
+
+        def check_termination_flag(termination_flag):
+            input("Press enter to exit the process.")
+            termination_flag.value = True
```
sneakers-the-rat, line 524: yep, let's use an `Event` :)

sneakers-the-rat (dismissed):
ok great! yes we're starting to see some of the changes that we'll need to make this testable. having sample input data and expected output data is great, that's enough to write a mock for the OKDev class. 

We do want to think about interface design here - in prior PR we punted on breaking up the config model into submodels, but i think we're starting to see here why that would be a good idea as we start adding in a bunch of new values to a flat model. User shouldn't need to be exposed to all the testing information, and that should all happen in a mock anyway. The purpose of having a hardware mock rather than special casing testing in the class itself is because in the latter case we are not testing the class itself, but instead the special branches that correspond to testing - ie. it's extremely easy (basically inevitable) for the main class to diverge in functionality and have that not be captured by the special testing code. it also allows us to encapsulate complexity so this class can stay simple - basically all testing stuff should not touch main package

sneakers-the-rat:
since the PR comes from a fork and not from a branch I can't push to it afaik, but see here: https://github.com/Aharoni-Lab/miniscope-io/commit/615eb344ec6e27ec4b0e69dd191d70b56ba7090f

- uses a mock instead of modifying class: this keeps the test consistent with in the wild behavior and also requires less special-casing. 
- rather than including the output file, we can instead hash the file created during tests and compare it against a known set of hashes - videos created on different platforms will be different from one another, and so this way we can keep the repo small and also handle those cases
- I also split out the video writer stuff into its own method and removed from config - we don't necessarily want to configure *everything* from the `yaml` file, and that seems like a pretty good example: if we wanted to split up the model and have a separate `Capture` submodel that could hold those things, then that would probably be fine, but we also want to avoid these models turning into "junk drawers" where we just keep adding fields indefinitely. similarly to previous model refactors, other devices will probably want to have video output, and we should create a common set of tools for that instead of each having their own.
- removed some zombie code as I came across it

sneakers-the-rat:
we'll definitely want to redo these queues because they seem to be pretty slow - i would expect processing 118 frames to be more or less instantaneous but it takes ~15s (longer than the video), so that's def a bottleneck for real usage

t-sasatani:
Thanks @sneakers-the-rat for the review and changes! I merged your branch and also added you to the fork in case. I'll look through the comments and address remaining issues within a few days.

sneakers-the-rat:
there's a lil more that could be done but ya i think it's starting to take shape! time to start making some structural changes to start moving it towards the kind of user interface/usage patterns we want for an sdk like this :)

t-sasatani:
I was curious why `multiprocessing.set_start_method('fork')` passes tests on Windows and noticed all tests are now on Linux. It might be an option to narrow down for now.

t-sasatani:
This might be something stupid but I'm confused why `conftest.py` doesn't seem to patch `okDev` when running pytest locally.

> (miniscope-io-py3.11) PS C:\Users\Takuya\GitHub\miniscope-io> pytest .\tests\test_stream_daq.py -v --log-cli-level=DEBUG
> =============================================== test session starts ===============================================
> platform win32 -- Python 3.11.6, pytest-7.4.4, pluggy-1.5.0 -- C:\Users\Takuya\GitHub\miniscope-io\.venv\Scripts\python.exe
> cachedir: .pytest_cache
> rootdir: C:\Users\Takuya\GitHub\miniscope-io
> configfile: pytest.ini
> plugins: anyio-4.4.0, timeout-2.3.1
> collected 1 item
> 
> tests/test_stream_daq.py::test_video_output[stream_daq_test_200px.yml-stream_daq_test_fpga_raw_input_200px.bin-hash0] +++++++++++++++++++++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++++++++++++++++++++
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Captured stderr ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
> Process Process-2:
> Traceback (most recent call last):
>   File "C:\Users\Takuya\AppData\Local\Programs\Python\Python311\Lib\multiprocessing\process.py", line 314, in _bootstrap
>     self.run()
>   File "C:\Users\Takuya\AppData\Local\Programs\Python\Python311\Lib\multiprocessing\process.py", line 108, in run   
>     self._target(*self._args, **self._kwargs)
>   File "C:\Users\Takuya\GitHub\miniscope-io\miniscope_io\stream_daq.py", line 241, in _fpga_recv
>     dev = okDev()
>           ^^^^^^^
>   File "C:\Users\Takuya\GitHub\miniscope-io\miniscope_io\devices\opalkelly.py", line 25, in __init__
>     raise DeviceOpenError("Cannot open device: {}".format(serial_id))
> miniscope_io.exceptions.DeviceOpenError: Cannot open device:

sneakers-the-rat:
> noticed all tests are now on Linux.

Ack good catch!!!

Yes monkeypatching something in a separate process is a little tricky and it depends on how the environment is created. I think I might need to add another method that hooks into the import machinery bc basically the behavior we want is "at test time, in all cases swap out okDev with the mock" even when a new process is spawned rather than forked

sneakers-the-rat:
You're probably gonna hate me for this but I think we should fix the process termination in the capture method instead of adding a timeout in tests 🙈.

I think it'll be worth it - basically make a shared `Event` that controls the loop for each of the subroutines, and if any of them throws an exception they can flip the event and all exit cleanly.

More generally I think we may want to take those out of processes and just turn them back into a single linear loop for now, bc we dont know exactly what should be parallelized yet without careful profiling, as is two of the processes are bottlenecked and might as well be done in serial. That will probably help with complexity as we get this thing on its feet, and we can add multiprocessing back in once we know where we need perf

t-sasatani:
Yes, I think we need `Event` for normal-ish operation anyway, and I think I'll have time to work on it this weekend, but I mentioned timeouts for something that kills the process from outside when even these get stuck.

> I think we may want to take those out of processes and just turn them back into a single linear loop
We can think about this later, but for now, I'll just look into apparent causes because I don't think I can make time to experiment for a while.

If monkeypatching becomes complex with multiprocessing, I think we can also limit to fork and UNIX-based systems for now. (I honestly think we can even temporarily move forward by directly plugging in a mock instance, though I know you'll hate that.)

sneakers-the-rat:
> We can think about this later

Agreed. Lets make this pass, clean it up a bit, merge, and try that in a separate PR.

> directly plugging in a mock instance

doing a 

```
if os.environ.get("PYTEST_VERSION") is not None:
    dev = okDevMock()
else:
    dev = okDev()
```

would be an OK, temporary hack while we work out multiprocessing - the main thing is to avoid making it so the test conditions are substantially different than the runtime conditions, which that wouldnt really do, and secondarily to keep the code tidy by not having too much testing stuff in the package itself. I wouldnt want it to stick around too long bc we have other devices to implement and we will want a generalizable mocking pattern for testing those too, but if we cant figure out how to monkeypatch in a spawned process then I think thats a bit less fragile than trying to force an OS to use a poorly supported mp mode.

t-sasatani:
I made some small updates because I had a minute. Now, this pytest module passes on my local Windows PC but not yet on GH (Linux: assert in stream DAQ test isn't passing; Win/Mac: many other failures too.). I'll work on the rest after tomorrow.
- Made the test code generate reference data hashes because that seems less fragile and also convenient. Please revert if we shouldn't do this.
- Plugged in the Mock module directly. Also, updated the reference video because having the read length argument in the mock changes the output a bit.

sneakers-the-rat:
> Made the test code generate reference data hashes because that seems less fragile and also convenient. Please revert if we shouldn't do this.

The purpose of using the hash is to not have to keep a reference video in the repo, just the hashes - since the hash of the video changes based on platform (video codec differences and etc.) Having several hashes that the generated video can match is one way to accomodate that. To avoid that we would probably want to encode the video as raw which would also avoid compression inconsistencies.

sneakers-the-rat:
hang on i've got this working by hashing the decoded video rather than the file, since the file will probably always have some variability both across platforms and runs.

sneakers-the-rat:
bear with me lol we can get this working - we'll want this for later bc we'll eventually have a lot of end to end tests and we don't want to have like 100MB of videos in the repo

t-sasatani:
Thanks, Jonny. I had no idea there could be ambiguity between runs. I thought having uncompressed videos would just approximately double the FPGA RAW data amount, so we could get away with that worst-case scenario, but I guess that wouldn't work if we're doing longer mock data generation and testing. Also, I'm curious how generating hashes for multi-platform and enough runs can be made easy.

Anyways I'll hold off for a while.

sneakers-the-rat:
ok sorry for the big squash commit there from https://github.com/Aharoni-Lab/miniscope-io/tree/tests_hash_video - re-activating the windows tests (thank you!!!!) exposed a bunch of path errors that were sort of separate from anything we were doing here. fixed those.

So what i did is i made a thing to use the decoded video as the basis for the hash, rather than the encoded bitstream of the video file. that seems to have evened out one of the sources of inconsistency. but it also seems like divx and x264 are both nondeterministic in general, so we just can't use them when we do video equality tests. we can have separate tests for codec support if we need them.

miniscope_io/models/stream.py
```diff
@@ -68,23 +72,33 @@ class StreamDaqConfig(MiniscopeConfig, YAMLMixin):
...
+    save_video_name : str, optional
+        If video is saved, , by default False.
+    show_video : bool, optional
+        Whether the video is showed in "real-time", by default True.
 
     ..todo::
         Takuya - double-check the definitions around blocks and buffers in the firmware and add description.
     """
+
+    mode: Literal["DAQ", "RAW_RECORD", "RAW_REPLAY"] = 'DAQ'
```
sneakers-the-rat, line 82: ok my last question is about this parameter - i get the reasoning for this, do we think this is still useful after using the mock? the `REPLAY` is no longer used, and what do you think about replacing the `RAW_RECORD` functionality by having a flag in the `capture` method like `raw: Optional[Path] = None` to enable recording the raw binary stream? 

i think we also probably want to add some controls for capturing only a set number of frames (or else make an iterator where we can do something like `for frame in daq.iter(15)` etc. as you have done when capturing for the record mode, but maybe save for another pr?

miniscope_io/stream_daq.py
```diff
@@ -237,7 +243,17 @@ def _fpga_recv(
...
             pre = pre[::-1]
         while True:
-            buf = dev.readData(read_length)
+            try:
+                buf = dev.readData(read_length)
+            except (EndOfRecordingException, StreamReadError):
+                self.terminate.value = True
+                break
+
+            if self.config.mode == 'RAW_RECORD':
```
sneakers-the-rat, line 252: flagging this as the other place we would need to modify if we went with suggestion ^

tests/data/stream_daq_test_output_200px.avi
sneakers-the-rat, line 1: if we are good with using hashes instead of the videos, i think we can delete this file?

sneakers-the-rat (commented):
OK, i think this is good to go, but have two remaining questions here, one on the `mode` parameter, and the other on the video file. thx for sticking with me through this PR, i think this is a big step forward!

sneakers-the-rat:
i think technically this should be a minor version bump but we've been using patch for units of work this size... works for me, maybe in `0.2` we start incrementing minor versions ;)

t-sasatani:
Thanks for all of the updates! I'm just skimming through the code, but it was nice to learn stuff like how you handled this ambiguity with the `utils.hash_video` method.

You know best about versioning, so whatever is good for you is good. The last few updates advanced the streamDAQ part pretty much, so it makes sense from my end to call this a `0.2`.

miniscope_io/models/stream.py
t-sasatani: Making this a flag of the capture method makes sense. I first did mode because we'll want stuff like bit error check mode, etc. that users will use. However, I guess adding flags like `output: Literal["video", "raw", "BER"]` will be better once we have some cli commands that internally configure this. (Either work but I can't imagine a scenario that needs both video and RAW, so we might not need an independent flag for this.)

An iterator makes sense because I was manually changing read length, which doesn't make sense as a workflow. I guess we can do this in close future PR.

tests/data/stream_daq_test_output_200px.avi
t-sasatani: I think it's better to keep it for now because knowing file sizes, how the video looks, etc., will help locate the problem when the test fails. I think it'll be best if we can host this somewhere outside of git (or an LFS server) in the future, but just for now.

miniscope_io/stream_daq.py
t-sasatani: I wasn't sure how these worked and misunderstood that adding the specified USB driver made the test go through, but I noticed these are just software modules, and that's not the case.

t-sasatani:
Also, I can test the operation with hardware in case by early next week, but I guess we can merge this before that, and it can be a `0.2.1` if we need some kind of fix.

miniscope_io/models/stream.py
sneakers-the-rat: ok did that here: https://github.com/Aharoni-Lab/miniscope-io/pull/26/commits/60221fa384bdedd94be79a752fe4e5030d5c7292

i think we'll probably want to do a more general overhaul of the notion of "sinks" or capture methods - depending on the use case, someone might want to iterate and grab any frames as fast as they can, otherwise they might want to pull from a queue to be sure they're not missing any, direct write to a video file in noninteractive mode, and so on, so i think we can save that for later and call this good for now :)

sneakers-the-rat:
ok fixed some few errant bugs, resolve the last changes, i think i'm gonna merge and publish this!!!!!

PR #31, opened 2024-06-26
Add coveralls

sneakers-the-rat:
Add coveralls test coverage reporting to enforce new code has tests (and also give us a target of what needs testing)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,9 +3,17 @@ name: Run Python tests
 on:
   push:
   workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+
+# Don't run duplicates - eg. if pushing to an open PR
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
 
 jobs:
-  build:
+  test:
     name: Run tests
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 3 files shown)


sneakers-the-rat:
OK once we get coveralls app added to the lab organization we should get PR comments back from coveralls, but for now you can see what this does at https://coveralls.io/github/Aharoni-Lab/miniscope-io

what ya think @t-sasatani ?

daharoni:
OK. I think I just added the Coveralls app. Let me know if it isn't working.

tests/test_stream_daq.py
```diff
@@ -4,19 +4,20 @@
...
-            '40047689185bdbeb81829aea7c6e3070bcd4673976e5836a138c3e1b54d75099'
+            "stream_daq_test_200px.yml",
+            "stream_daq_test_fpga_raw_input_200px.bin",
+            "40047689185bdbeb81829aea7c6e3070bcd4673976e5836a138c3e1b54d75099",
         )
-    ]
+    ],
 )
-@pytest.mark.timeout(30)
+@pytest.mark.timeout(60)
```
t-sasatani, line 18: This is a bit unrelated, but any test could theoretically get stuck, so should we just have something global in `pyproject.toml` instead? It's very minor, but I was wondering because you changed this and the decorator below.
```
[tool.pytest.ini_options]
timeout = "60"
```

t-sasatani (approved):
I didn't know this kind of thing existed, and it looks useful. Let's merge this before #29. (I'll address your reviews in that one by early next week.)

tests/test_stream_daq.py
sneakers-the-rat: that is not a bad idea to just have a global timeout - i mean yeah i would consider any test taking longer than 60 seconds a failure for sure.

PR #48, opened 2024-10-24
Over The Air device config feature

t-sasatani:
This PR adds modules and CLI to update device parameters over the air.

**To-dos**

Add more commands
- [x] Gain
- [x] FOV

Tests
- [x] Command generation
- [x] Models and validation
- [x] Mock device I/O (though not sure what's the best way because this is one-way communication)

Others/low priority
- [x] Documentation
- ~Integrate with the capture command~(Future work)
- ~Check metadata and validate device update (plotting and checking by eye might be good for now)~(Future work)


sneakers-the-rat:
hell ya

t-sasatani:
Yup will ping you in a bit!

t-sasatani:
@sneakers-the-rat I'd appreciate it if you could have a look. I'm mostly unsure what a good way to test device I/O is.

t-sasatani:
Oh last commit I pushed broke some import. very easy fix so I'll doit when im back on pc.

miniscope_io/cli/update.py
```diff
@@ -0,0 +1,77 @@
...
+    "-i",
+    "--device_id",
+    required=False,
+    default=0,
+    type=int,
+    help="ID of the device to update. 0 will update all devices.",
+)
+@click.option(
+    "-t",
+    "--target",
```
sneakers-the-rat, line 29: maybe "key"? (if this is a key/value pair?)

miniscope_io/device_update.py
```diff
@@ -1,144 +1,105 @@
...
+        time.sleep(0.1)
+
+        serial_port.write(UpdateCommandDefinitions.reset_byte.to_bytes(1, "big"))
+
+    finally:
+        serial_port.close()
+        logger.info("Closed serial port")
+
+
+def find_ftdi_device() -> list:
```
sneakers-the-rat, line 92:
```suggestion
def find_ftdi_device() -> list[str]:
```

miniscope_io/device_update.py
```diff
@@ -1,144 +1,105 @@
...
+    """
+    Find FTDI devices connected to the computer.
+    """
```
sneakers-the-rat, line 95:
```suggestion
    """
    Find FTDI devices connected to the computer.
    
    Returns:
        list[str]: list of matching device paths, e.g. `/dev/ttyUSB0` from :attr:`~pyserial.tools.list_ports.ListPortInfo`
    """
```

miniscope_io/models/devupdate.py
```diff
@@ -0,0 +1,123 @@
...
+    MSB_header = 0b10000000
+    LSB_value_mask = 0b000000111111  # value below 12-bit
+    MSB_value_mask = 0b111111000000  # value below 12-bit
+    reset_byte = 0b11111111
+
+
+class UpdateTarget(Enum):
+    """
+    Targets to update. Needs to be under 6-bit.
+    """
```
sneakers-the-rat, line 36: we will eventually want to unify this with a device configuration, since I assume these values won't be static for all miniscopes for all time - so e.g. we have a common set of targetable values and then each device provides a mapping to the int code/a method for setting each of them. 

fine for now tho, since the ABC refactor should actually give us a reasonable place to put these things

miniscope_io/models/devupdate.py
```diff
@@ -0,0 +1,123 @@
...
+        elif target == UpdateTarget.DEVICE:
+            assert value in [DeviceCommand.REBOOT.value], "For DEVICE, value must be in [200]"
+        elif (
+            target == UpdateTarget.ROI_X
+            or target == UpdateTarget.ROI_Y
+            or target == UpdateTarget.ROI_WIDTH
+            or target == UpdateTarget.ROI_HEIGHT
+            or target == UpdateTarget.EWL
+        ):
+            pass  # Need to implement
```
sneakers-the-rat, line 81:
```suggestion
        elif target in UpdateTarget:
            raise NotImplementedError()
        else:
            raise ValueError(f"{target} is not a valid update target, need an instance of UpdateTarget")
```

sneakers-the-rat: I see the invalid target validator below, adding the else clause for logical completeness

miniscope_io/models/devupdate.py
```diff
@@ -0,0 +1,123 @@
...
+    MSB_value_mask = 0b111111000000  # value below 12-bit
+    reset_byte = 0b11111111
+
+
+class UpdateTarget(Enum):
+    """
+    Targets to update. Needs to be under 6-bit.
+    """
+
+    LED = 0
```
sneakers-the-rat, line 38: using enums as mappings to meaningful values like this is usually more trouble than an enum with opaque values and a separate value mapping (e.g. needing to get the value in the validator), but don't want to be too nitpicky. fine with me just giving my experience :)

miniscope_io/models/devupdate.py
```diff
@@ -0,0 +1,123 @@
...
+    id_header = 0b00000000
+    target_header = 0b11000000
+    LSB_header = 0b01000000
+    MSB_header = 0b10000000
+    LSB_value_mask = 0b000000111111  # value below 12-bit
+    MSB_value_mask = 0b111111000000  # value below 12-bit
+    reset_byte = 0b11111111
+
+
+class UpdateTarget(Enum):
```
sneakers-the-rat, line 33:
```suggestion
class UpdateTarget(int, Enum):
```

if you do this, then the members of the enum behave like integers as well as enum members:

```python
class RegularEnum(Enum):
    x = 1
    y = 2
    z = 3
    
class IntEnum(int, Enum):
    x = 1
    y = 2
    z = 3
```

```python
>>> isinstance(RegularEnum.x, int)
False
>>> RegularEnum.x + 5
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'RegularEnum' and 'int'
>>> RegularEnum.x in RegularEnum
True
>>>
>>> isinstance(IntEnum.x, int)
True
>>> IntEnum.x + 5
6
>>> IntEnum.x in IntEnum
True
```

miniscope_io/models/devupdate.py
```diff
@@ -0,0 +1,123 @@
...
+        Returns:
+            UpdateTarget: Validated target as UpdateTarget.
+
+        Raises:
+            ValueError: If target not found.
+        """
+        try:
+            return UpdateTarget[value]
+        except KeyError as e:
+            raise ValueError(f"Target {value} not found.") from e
```
sneakers-the-rat, line 123:
```suggestion
            raise ValueError(f"Target {value} not found, must be a member of UpdateTarget: {list(UpdateTarget.__members__.keys())}.") from e
```

tests/test_device_update.py
```diff
@@ -0,0 +1,160 @@
...
+@patch("miniscope_io.device_update.serial.Serial")
+def test_devupdate_with_device_connected(mock_serial, mock_comports):
+    """
+    Test DevUpdate function with a device connected.
+    """
+    mock_serial_instance = mock_serial.return_value
+    mock_comports.return_value = [
+        MagicMock(vid=0x0403, pid=0x6001, device="COM3"),
+        MagicMock(vid=0x1234, pid=0x5678, device="COM4"),
+    ]
```
sneakers-the-rat, line 19: can we put this in a fixture? or is there a reason that we're mocking in a test function?

sneakers-the-rat: but also since i know that you have experiments to run in the near future, i'd be ok pushing this off if we just raise an issue to remind ourselves to do it

sneakers-the-rat (approved):
Some suggestions on error messages, but overall i'd be fine with merging this as long as we raise an issue to remind us to write a full mock of the Serial device later. I would really prefer we stick to python naming conventions and use snake case for function names, but don't want to be a dictator about it.

Other than that, hopefully comments are self explanatory

miniscope_io/device_update.py
sneakers-the-rat: just pushed a commit to add pyserial to the intersphinx mapping so that this works.

sneakers-the-rat:
oh crap sorry i hit rebase and forgot that it adds me as co-author on all the commits :'\. should have waited for u

t-sasatani:
Thanks for the comments! I'll look into it in a bit. And the commits are no problem; I'm not sure if I can interact with origin this week anyway.

miniscope_io/cli/update.py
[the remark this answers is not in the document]
t-sasatani: Yeah, the key makes more sense. I will update it.

miniscope_io/cli/update.py
```diff
@@ -0,0 +1,77 @@
...
+    required=False,
+    type=int,
+    help="Value to set. Must be used with --target and cannot be used with --restart.",
+)
+@click.option(
+    "-c",
+    "--config",
+    required=False,
+    type=click.Path(exists=True, dir_okay=False),
+    help="YAML file with configuration to update. Specify target and value pairs in the file.",
```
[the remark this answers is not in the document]
t-sasatani: I separated this because the device config is static unless the firmware is updated, but the user will likely interactively change this update stuff while looking at the data, and having them together felt awkward.
It's just a lazy batch interface that literally sends the YAML file's keys/value pairs like `mio update --key KEY --value VALUE`.

This batch stuff isn't strictly necessary, and I'm not sure what's the best way to have this, so I'd like to keep it and just note this is an experimental feature in the help if that sounds ok.

miniscope_io/cli/update.py
```diff
@@ -0,0 +1,77 @@
...
+)
+@click.option(
+    "--reboot",
+    is_flag=True,
+    type=bool,
+    help="Restart the device. Cannot be used with --target or --value.",
+)
```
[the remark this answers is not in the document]
t-sasatani: Yeah I was just lazy for this and makes sense to separate.

miniscope_io/cli/update.py
```diff
@@ -0,0 +1,77 @@
...
+    "-p",
+    "--port",
+    required=False,
+    help="Serial port to connect to. Needed if multiple FTDI devices are connected.",
+)
+@click.option(
+    "-i",
+    "--device_id",
+    required=False,
+    default=0,
```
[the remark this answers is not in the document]
t-sasatani: It's just a number burned into the firmware and nothing OS-related (similar to serial number in SCPI instruments). It's used to filter out commands sent to a specific Miniscope when there are several scopes in one experiment.

https://github.com/Aharoni-Lab/Miniscope-SAMD-Framework/blob/13fbb85c114ac290a462bda1250c6406384c9acb/src/MS_callback.c#L119

miniscope_io/device_update.py
```diff
@@ -1,144 +1,105 @@
...
+    FTDI_VENDOR_ID = 0x0403
+    FTDI_PRODUCT_ID = 0x6001
```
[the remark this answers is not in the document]
t-sasatani: Probably not unless FTDI is in trouble.

miniscope_io/device_update.py
t-sasatani: Makes sense.

miniscope_io/models/devupdate.py
t-sasatani: Kind of depends on how we want the interface to look because these have quite different staticness. So, suppose there is a GUI; the current device config (and maybe initial values for this update) should be used for initiating, and the stuff in this update should be handled in the interface.

t-sasatani: But anyways let's leave it for later

tests/test_device_update.py
t-sasatani: Oh yea I was playing around with COM/pids while writing this and just left it that way. fixtures will scale better.

t-sasatani: I was initially thinking about full mock, but this device literally has no response and accepts commands in one-way. The response comes through parameter changes coming in from the `streamdaq` module, and I don't have time to integrate these so ended up this way. But makes sense to keep it as an issue in mind

t-sasatani: not mind probabl ygithub

miniscope_io/device_update.py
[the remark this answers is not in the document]
sneakers-the-rat: cool cool. only think i would suggest is moving constants to module level as a matter of style consistency, but idrc since i am not sure when else they would be needed and we don't have a code organization system where it would be obvious for someone to go looking for FTDI constants.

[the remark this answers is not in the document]
t-sasatani: OK, I'd usually try to think about putting specific variables in specific functions to save the namespace (and end up putting everything in global scope lol), but I don't really care for this case.

miniscope_io/models/devupdate.py
t-sasatani: I'm not sure if I get your intention here. Would you mind making this change so I can see through your example?

miniscope_io/models/devupdate.py
t-sasatani: Oh this is nice

miniscope_io/cli/update.py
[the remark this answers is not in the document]
sneakers-the-rat: got it! made a note to document this later

miniscope_io/cli/update.py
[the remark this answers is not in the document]
sneakers-the-rat: > the user will likely interactively change this update config stuff while looking at the data, and having them together felt awkward.

def true. to make them be the same thing would require us making this update function idempotent and good at filtering out irrelevant values and unchanged values. fine to keep as is, think the note would be helpful so that ppl know this is a different file than what we mean when we use "config" elsewhere (we might even just name it "yaml" to say "pass k/v pairs via yaml" rather than call it a "config" to avoid confusion, but idrc about that)

miniscope_io/device_update.py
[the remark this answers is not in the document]
sneakers-the-rat: definitely agreed globally scoped variables bad, and it's unfortunate there are not true consts in python. allcaps naming is not much of a protection ;)

miniscope_io/cli/update.py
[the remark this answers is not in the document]
t-sasatani: doen

miniscope_io/cli/update.py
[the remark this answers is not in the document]
t-sasatani: done

miniscope_io/device_update.py
```diff
@@ -1,144 +1,105 @@
...
-updateDeviceParser = argparse.ArgumentParser("updateDevice")
-updateDeviceParser.add_argument("port", help="serial port")
-updateDeviceParser.add_argument("baudrate", help="baudrate")
-updateDeviceParser.add_argument("module", help="module to update")
-updateDeviceParser.add_argument("value", help="LED value")
+logger = init_logger(name="device_update", level="INFO")
 
 
-def updateDevice() -> None:
+def DevUpdate(
```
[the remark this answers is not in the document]
t-sasatani: done

miniscope_io/models/devupdate.py
t-sasatani: done

miniscope_io/models/devupdate.py
sneakers-the-rat: it can be somewhat fragile to treat the value of an enum as meaningful, at least in python - i think a good rule of thumb is to treat them like you would RDBMS `id` columns or persistent identifiers, opaque unique values that identify an object but don't contain information so that the information can freely change without the identifier needing to change.

thinking about this situation recently: https://github.com/linkml/linkml-runtime/pull/310#issuecomment-2011023357

where the value of an enum was meaningful, and then over time it became the case accidentally that multiple enum keys got the same value (which doesn't work and invisibly caused problems) and it wasn't really possible to reuse those enum keys and give them further information as became necessary.

so the idea of an enum is good here, we do want to have some representation of the possible kinds of commands that can be given, and i think it would be even better if this was an enum for all possible kinds of commands across all miniscopes. that might not have been what you had in mind here, but we will need something like that eventually and this seems as good a candidate as any.

one might imagine wanting to do something like this:

```python
WirelessMiniscope.configure(UpdateTarget.GAIN, 10)
WirelessMiniscopeV2.configure(UpdateTarget.GAIN, 10)
AnotherMiniscope.configure(UpdateTarget.GAIN, 10)
EphysDevice.configure(UpdateTarget.GAIN, 10)
I2CDevice.configure(UpdateTarget.GAIN, 10)
# ... and so on
```

if the value of the enum is meaningful beyond being a unique identifier though, then...
- we wouldn't be able to e.g. modify the firmware and use a different value, 
- or we would have to be careful to ensure that the value was the same across all the firmwares of all devices, 
- or we would have to have mappings between the enum value and the specific value that some different firmware uses to set GAIN
- we would have trouble adapting it to different modalities like i2c where the value might be some hex register address

we would also sacrifice some ergonomics that might be nice to have, e.g. we might want to allow someone to just pass a string like "`miniscope.set('gain', 10)`" instead of needing to import the enum (while also constraining the possible values that can be accepted, so not just having the signature of `set` be `set(key: str, ...)` but instead be something like `set(key: SomeKeyEnum, ...)`). 

So we might do something like this:

```python
class UpdateTarget(str, Enum):
    # behaves similarly to inheriting from int and enum, 
    # part of stdlib `enum` as of 3.11
    LED = "led"
    GAIN = "gain"
    # ...
    
class SomeMiniscope(Device):
    command_map: ClassVar[dict[UpdateTarget, int]] = {
        UpdateTarget.LED: 0,
        UpdateTarget.GAIN: 1,
        # ...
    } 
    
    def set(self, key: UpdateTarget, value: Any):
        # say this is some hardware abstraction class like OkDev...
        self._inner_hardware.set(self.command_map[key], value)

```

where the value of the enum is still useful as a string, but carries no further information aside from being one of a set of unique values. 

again, doesn't have to be in this PR, and also this might not be what this class is for, but just explaining what i was thinking in previous comment

miniscope_io/cli/update.py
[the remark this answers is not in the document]
t-sasatani: This is basically a batch so I will switch to `yaml_batch`. Not beautiful so pretty temporary.

tests/test_device_update.py
sneakers-the-rat: > I was initially thinking about full mock, but this device literally has no response and accepts commands in one-way

hahaha ok, that makes sense. silent device makes for uninteresting mocks. yeah let's raise it as an issue to track needing to do once the system is more mature/we're closer to release and we want to do formal verification. a fixture that bundles together monkeypatching the relevant methods/attrs like you have already done here would be nice, as you say, for scale, but understood why simulating the device behavior here might be less than enlightening :). 

something like this that just stashes the values passed might be nice given the pattern of some of these tests? https://github.com/Aharoni-Lab/miniscope-io/blob/56b32b7345698c004c1170aac664276ff7836de7/miniscope_io/devices/mocks.py#L74

but also w/e is fine.

miniscope_io/models/devupdate.py
```diff
@@ -74,7 +74,9 @@ def validate_values(cls, values: dict) -> dict:
         elif target in UpdateTarget:
             raise NotImplementedError()
         else:
-            raise ValueError(f"{target} is not a valid update target, need an instance of UpdateTarget")
+            raise ValueError(
+                f"{target} is not a valid update target," "need an instance of UpdateTarget"
```
sneakers-the-rat, line 78: this happens to me literally all the time where i want to split a line for aesthetic reasons and then black is like "no you should not have split that line." it borders on malicious compliance and i imagine `black` having like a smile on its face every time it does this like "you were the one who set the line width!"

miniscope_io/models/devupdate.py
t-sasatani: Ah, I get your point. Thanks for the detailed guide.
It makes sense that using `enum` this way is vulnerable to accidents as we extend this in multiple ways, but it also still feels pretty arbitrary regarding how we should implement this to me, so I will skip it for now.

miniscope_io/models/devupdate.py
t-sasatani: Oh, I was fully trusting `black` and wasn't even paying attention to the changes it made lol.

miniscope_io/models/devupdate.py
sneakers-the-rat: that is fair. it is maybe premature to think about uniformity of interface vs. heterogeneity of implementation when we don't yet have a uniform interface.

miniscope_io/models/devupdate.py
sneakers-the-rat: i think it is literally equivalent to the interpreter after parsing (?), but at the same time there is something bad about `("start " "end")` vs `"start end"` that feels like the machines are taking revenge when they make it happen.

tests/test_device_update.py
t-sasatani: Makes sense, but I think I'll skip this for now as there's much higher priority stuff in my task queue. I'll just change to parameterized fixtures for this PR.

sneakers-the-rat: github needs a better interface for "just talking, not trying to hold up work, merge whenever" bc that's where i've been at for awhile on this one

t-sasatani: Yea I was going to merge whenever I do the fixture stuff, so not really held by your comments, but there could be a chat interface.

PR #50, opened 2024-11-05
switch to PDM, rm nox

sneakers-the-rat:
Poetry and I have fallen out of love, and it will especially hold us back once we want to start doing compiled extension modules. 

Should be no functional change here, and wherever you were using poetry before just swap it out to pdm.

e.g. 
```shell
# install 
pdm install
# or (from a venv)
pip install -e .

# run `mio` cli
pdm run mio
# or, if installed in an activated venv, still just
mio

# run tests
pdm run test
# lint
pdm run lint
# format
pdm run format

# add dependency
pdm add package_name
```


t-sasatani (approved):
Looks good to me, and makes sense to make this switch.

PR #53, opened 2024-11-05
[ci] Add readthedocs preview links to PRs

sneakers-the-rat:
I'm gonna :self-high-five: this one, all it is doing is adding a CI action that adds a link to the readthedocs build for a PR to the OP issue when a PR is opened:

see:
https://github.com/readthedocs/actions/tree/v1/preview


t-sasatani (approved):
This can't cause problems but looks useful, so quickly approving before the big one.

Issue #59, opened 2024-11-07
[pipelines] Convert SDCard to `Camera` API


sneakers-the-rat:
Convert `miniscope_io.io.SDCard` to a `SDMiniscope` subclass of `Camera`

- [ ] Create `SDCard` source
    - [ ] Parent class `FileSource`
    - [ ] Designate sources as "push" vs "pull" to differentiate from streaming sources - streaming sources "push" data constantly as it is produced, "pull" sources assume the data is static and request it as the consumers need it
- [ ] Create `Generator` sink to be able to iterate over the outputs of a pipeline 
- [ ] Implement abstract methods on `Camera` and Device
- [ ] Adapt Config
- [ ] Update notebooks

PR #68, opened 2024-11-08
[pipelines] Convert SDCard to ABC and Pipeline system

sneakers-the-rat:
Fix: https://github.com/Aharoni-Lab/miniscope-io/issues/59

Opening a draft PR to run CI tests as i work on converting the SDCard miniscope to the pipeline system. This will be a sort of experimental PR as I go since i'm going to be figuring out what changes i need to make to the ABCs and whatnot to set us up for a general pipelining system that lets us accomodate all the different kinds of devices while also providing kind UX across different use patterns (e.g. the post-hoc SDCard miniscope has a very different ux than the online streaming scopes)

----
📚 Documentation preview 📚: https://miniscope-io--68.org.readthedocs.build/en/68/


sneakers-the-rat:
ok since all the changes up until here are just moving things around and repairing the `pipeline` branch after merging in main, i'm going to merge this to `pipeline` so that we can actually compare the substantive changes in the PR

PR #72, opened 2024-11-12
Global and User config, CLI config

sneakers-the-rat:
This PR ports some stuff i have worked on in linkml over to miniscope-io.

Specifically we

- Adds a configurable `user_dir` source of config - currently there is a place to set `base_dir` but no way to get config from that directory, but now one can define global config there too.
- Adds other sources of config because why the hell not - currently one can use ` .env` file, but this adds the ability to use (in order of priority, high to low)...
  - `mio_config.yaml` in current directory
  - `pyproject.toml` - the `[tool.miniscope_io.config]` table
  - `mio_config.yaml` in a custom user directory
  - `mio_config.yaml` in a fixed global directory 
- Adds cli commands to
  - create/set a user directory
  - show user/global configs
  - show locations of user and global config files

This also improves testing for configs, making fixtures to set each of the types of settings sources.

I know i have been like maniacal about config on this project, but i swear it's not just a pointless yak shave: as i said before, configuration is going to be the main point of interaction that people have with `miniscope-io` since it's an SDK not a full-blown GUI app, it might get bundled in with other programs, and it's the way that we're exposing as how to control the operation of the program (somewhat tautologically...). So it needs to be tight and evolvable. 

This sets us up to make the configuration of devices more straightforward, where we might have something like `mio device {device_type} create` to create a default configuration for that device into `{user_dir}/devices` so that they can then customize it. being able to have flexible configuration from the system level down to individual projects (i.e. global config vs. the directory-specific configs) then we make it possible to bundle device configs along with experimental data, so one could just cd to a different directory and automatically use the configs for the project in that directory.

----
📚 Documentation preview 📚: https://miniscope-io--72.org.readthedocs.build/en/72/

### DIFF
--- a/docs/guide/config.md
+++ b/docs/guide/config.md
@@ -1,0 +1,157 @@
+# Configuration
+
+```{tip}
+See also the API docs in {mod}`miniscope_io.models.config`
+```
+
+Config in `miniscope-io` uses a combination of pydantic models and
+[`pydantic-settings`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/).
+
+Configuration takes a few forms:
+
+- **Global config:** control over basic operation of `miniscope-io` like logging,
+  location of user directories, plugins, etc.
+- **Device config:** control over the operation of specific devices and miniscopes like
+  firmware versions, ports, capture parameters, etc.
+- **Runtime/experiment config:** control over how a device behaves when it runs, like
+  plotting, data output, etc.
+
+## Global Config
+
+Global config uses the {class}`~miniscope_io.models.config.Config` class
+
+Config values can be set (in order of priority from high to low, where higher
+priorities override lower priorities)
+
+* in the arguments passed to the class constructor (not user configurable)
+* in environment variables like `export MINISCOPE_IO_LOG_DIR=~/`
+* in a `.env` file in the working directory
+* in a `mio_config.yaml` file in the working directory
+* in the `tool.miniscope_io.config` table in a `pyproject.toml` file in the working directory
+* in a user `mio_config.yaml` file in the user directory (see [below](user-directory))
+* in the global `mio_config.yaml` file in the platform-specific data directory
+  (use `mio config global path` to find its location)
+* the default values in the {class}`~miniscope_io.models.config.Config` model
+
+Parent directories are _not_ checked - `.env` files, `mio_config.yaml`, and `pyproject.toml`
+files need to be in the current working directory to be discovered.
+
+You can see your current configuration with `mio config`
+
+(user-directory)=
+### User Directory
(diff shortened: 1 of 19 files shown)


docs/guide/config.md
```diff
@@ -0,0 +1,157 @@
...
+
+* in the arguments passed to the class constructor (not user configurable)
+* in environment variables like `export MINISCOPE_IO_LOG_DIR=~/`
+* in a `.env` file in the working directory
+* in a `mio_config.yaml` file in the working directory
+* in the `tool.miniscope_io.config` table in a `pyproject.toml` file in the working directory
+* in a user `mio_config.yaml` file in the user directory (see [below](user-directory))
+* in the global `mio_config.yaml` file in the platform-specific data directory
+  (use `mio config global path` to find its location)
+* the default values in the {class}`~miniscope_io.models.config.Config` model
```
t-sasatani, line 34: Yes, making this explicit helps a lot.

docs/guide/config.md
```diff
@@ -0,0 +1,157 @@
...
+
+```{command-output} mio config user create --dry-run
+```
+
+### Setting Values
+
+```{todo}
+Implement setting values from CLI.
+
+For now, please edit the configuration files directly.
```
t-sasatani, line 77: It is also related to my comments in #76, but I'm not sure which is easier between these two (though it wouldn't hurt to have these two work nicely). I feel that using text editors could be easier than using CLI for our target users (~undefined), but I am unsure.

miniscope_io/cli/config.py
```diff
@@ -0,0 +1,148 @@
+"""
+CLI commands for configuration
+"""
```
t-sasatani, line 3: This looks very helpful.

t-sasatani (approved):
Looks very helpful! I was always confused around this.

sneakers-the-rat:
well i sorta merged these backwards and made this much harder for myself.... my bad

docs/guide/config.md
sneakers-the-rat: totally ya. after seeing you working with the cli setting config values i think that would be really nice to have, and that should also edit the relevant files, so they should be made to be equivalent imo if we do set config values from the cli - i.e. editing from cli should change the value in the config files as if one were to edit it with a text editor

sneakers-the-rat:
I think that having a separate `devices` configs directory is probably more complex than we want. having a single flat id-based system where the user can structure it into whatever they want within `configs` is probably easier than having to keep track of where files are stored. gonna remove that part (it's not used anyway in this PR)

sneakers-the-rat:
added some tests and we should be good to go

Issue #69, opened 2024-11-09
`MINISCOPE_IO_LOGS__LEVEL` in `.env` doesn't set log level


t-sasatani:
It seems like `MINISCOPE_IO_LOGS__LEVEL` in the `.env` file doesn't change the stdout/file log levels.
Below is the added test function and a GitHub action test run that should explain this.
https://github.com/Aharoni-Lab/miniscope-io/actions/runs/11756676032

t-sasatani:
Adding test for this here too in case.
```python
@pytest.mark.parametrize('level', ['DEBUG', 'INFO', 'WARNING', 'ERROR'])
def test_init_logger_from_dotenv(tmp_path, level):
    """
    Set log levels from dotenv MINISCOPE_IO_LOGS__LEVEL key
    """
    # Feels kind of fragile to hardcode this but I couldn't think of a better way so for now
    level_name_map = {
        'DEBUG': logging.DEBUG,
        'INFO': logging.INFO,
        'WARNING': logging.WARNING,
        'ERROR': logging.ERROR
    }

    tmp_path.mkdir(exist_ok=True,parents=True)
    dotenv = tmp_path / '.env'
    with open(dotenv, 'w') as denvfile:
        denvfile.write(f'MINISCOPE_IO_LOGS__LEVEL="{level}"')
    
    dotenv_logger = init_logger(name='test_logger', log_dir=tmp_path)

    assert dotenv_logger.level == level_name_map.get(level)
```

sneakers-the-rat:
The `.env` file needs to be in the current working directory, we sort of cheat that in the current test by passing the  .env file explicitly: https://github.com/Aharoni-Lab/miniscope-io/blob/fa69f0df8c66dbf9eb4b30690a5c815df98945b0/tests/test_config.py#L46

And I should update that now that ive written better tests elsewhere.

If you monkeypatch the current directory to that tmp dir it should work: https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.chdir

sneakers-the-rat:
Ill go ahead and add some stuff I did for linkml to have a user directory where you can set that globally in addition to current directory

https://github.com/linkml/linkml/pull/2335

t-sasatani:
I fixed the path stuff. Now, it should initialize the logger after moving to the `tmp_path` and test each handler separately.
I still might have bugs in the test, but it seems like the propagation to the handlers isn't working.
```python
[...]

PR #73, opened 2024-11-12
Single root logger handler :)

sneakers-the-rat:
Fix: https://github.com/Aharoni-Lab/miniscope-io/issues/69

Instead of having each logger having their own handler, we instead have one single root logger `miniscope_io` that has all the handlers, and then all the child loggers propagate messages up to it.

The biggest change in behavior here is that we now will have a single log file (or, rather, `n` log files depending on how many rotating files we have configured) rather than a log file per logger. If we prefer that, we can go back to that, but it requires us to also have multiple stdout handlers since we need to set `propagate = false` at the logger level to prevent double-logging, which we have mostly avoided by chance from not having any nested loggers.

Opening this as a draft for the moment because I think this will play screwy with multiprocessing, so let me test that real quick too.

edit: ok added multiprocessing handling - if a logger is spawned in a child process, it logs to a separate file that includes the logger's name as well as the PID (anticipating people wanting to run multiple of the same kind of miniscope on the same machine).

----
📚 Documentation preview 📚: https://miniscope-io--73.org.readthedocs.build/en/73/


miniscope_io/logging.py
```diff
@@ -48,27 +49,86 @@ def init_logger(
...
         log_file_n = config.logs.file_n
     if log_file_size is None:
         log_file_size = config.logs.file_size
 
+    # set our logger to the minimum of the levels so that it always handles at least that severity
+    # even if one or the other handlers might not.
+    min_level = min([getattr(logging, level), getattr(logging, file_level)])
+
     if not name.startswith("miniscope_io"):
         name = "miniscope_io." + name
```
t-sasatani, line 65: Not really related here, but would it hurt to change this to `mio`? `miniscope_io` is pretty long for my small terminal

miniscope_io/logging.py
```diff
@@ -48,27 +49,86 @@ def init_logger(
...
+            _file_handler(
+                name=f"{name}_{mp.current_process().pid}",
+                file_level=file_level,
+                log_dir=log_dir,
+                log_file_n=log_file_n,
+                log_file_size=log_file_size,
+            )
+        )
+        logger.addHandler(_rich_handler(level))
+        logger.propagate = False
```
t-sasatani, line 90: Oh, didn't know how to do this. Neat.

t-sasatani (approved):
I think it looks great and we can merge.

miniscope_io/logging.py
sneakers-the-rat: yep! good idea. will do rn

sneakers-the-rat: done in: https://github.com/Aharoni-Lab/miniscope-io/pull/73/commits/3624fd8aa9c6cdc98d8058bc9b7053d1c32d7628

PR #74, opened 2024-11-12
Fix annoying pydantic warnings

sneakers-the-rat:
Currently the test output is a total clusterfuck because of warnings emitted when writing the header data.

The warnings come because we're using `model_construct()` to create the header models, which skips the validation and coercion step, so we have numpy ints/floats instead of base python ints/floats. They serialize just fine, so it's safe to just ignore the warning there. 

Also fixed a warning from python thinking we're trying to escape ``'\['` instead of making a rich format string, so changed that to a raw string.

now there are no noisy warnings!


sneakers-the-rat:
this one is uncontroversial so i'm just gonna YOLO merge it

Issue #71, opened 2024-11-11
`sphinx-click` doesn't document click


t-sasatani:
The CLI documentation doesn't appear in `v0.5.0`, though `cli/main.rst` hasn't changed—just a note for returning later. 
- v0.5.0: https://miniscope-io.readthedocs.io/en/v0.5.0/cli/main.html
- v0.4.1: https://miniscope-io.readthedocs.io/en/v0.4.1/cli/main.html

sneakers-the-rat:
Yeah I noticed that too, it builds locally but I think i messed smth up in the docs. Will add doc tests

sneakers-the-rat:
oof wait i still broke this, reopening bc i need to convert .rst formatted click commands to myst

sneakers-the-rat:
this works now!

PR #75, opened 2024-11-12
Test docs

sneakers-the-rat:
Fix: https://github.com/Aharoni-Lab/miniscope-io/issues/71

To keep the documentation from being broken, let's test the docs - this action builds the docs and raises warnings as errors.

Error from documenting the CLI seems to have come from needing to have matplotlib installed to import anything, which shouldn't be the case - we want matplotlib to be optional bc it's heavy. So i changed that importerror to a check when one attempts to use the streamplotter instead

----
📚 Documentation preview 📚: https://miniscope-io--75.org.readthedocs.build/en/75/

### DIFF
--- a/.github/workflows/docs-test.yml
+++ b/.github/workflows/docs-test.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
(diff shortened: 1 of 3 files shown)


t-sasatani (approved):
Looks like this can't break anything. Also, it's technically tested in the `experiment-YYYYMMDD` branch, so I think it's good.

PR #80, opened 2024-11-27
Bugfixes for metadata export

t-sasatani:
- Make metadata export keys and values consistent by writing keys after getting the first header.
- Allow `float` value export in metadata (before it was all rounded to `int`).
- Update test data to include computed fields and float values in tests.

This should be a draft, but I made it a real PR for collecting test hashes.


t-sasatani:
@sneakers-the-rat, there is no rush at all for this. I just requested it before I forget.

PR #82, opened 2024-11-27
Bound stream computed field with physical limits

t-sasatani:
- Sets limits on the computed ADC voltage value.
- The computed voltage is set to zero if it exceeds the limits.
- This doesn't affect the raw metadata value.

----
📚 Documentation preview 📚: https://miniscope-io--82.org.readthedocs.build/en/82/

### DIFF
--- a/miniscope_io/models/stream.py
+++ b/miniscope_io/models/stream.py
@@ -38,12 +38,12 @@ class ADCScaling(MiniscopeConfig):
     battery_max_voltage: float = Field(
         10.0,
         description="Maximum voltage of the battery."
-          "Scaled battery voltage will be 0 if it is greater than this value",
+        "Scaled battery voltage will be 0 if it is greater than this value",
     )
     vin_max_voltage: float = Field(
         20.0,
         description="Maximum voltage of the Vin"
-          "Scaled Vin voltage will be 0 if it is greater than this value",
+        "Scaled Vin voltage will be 0 if it is greater than this value",
     )
 
     def scale_battery_voltage(self, voltage_raw: float) -> float:
--- a/miniscope_io/models/stream.py
+++ b/miniscope_io/models/stream.py
@@ -35,6 +35,16 @@ class ADCScaling(MiniscopeConfig):
         11.3,
         description="Voltage divider factor for the Vin voltage",
     )
+    battery_max_voltage: float = Field(
+        10.0,
+        description="Maximum voltage of the battery."
+          "Scaled battery voltage will be 0 if it is greater than this value",
+    )
+    vin_max_voltage: float = Field(
+        20.0,
+        description="Maximum voltage of the Vin"
+          "Scaled Vin voltage will be 0 if it is greater than this value",
+    )
 
     def scale_battery_voltage(self, voltage_raw: float) -> float:
         """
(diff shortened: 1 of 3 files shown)


sneakers-the-rat:
seems fine to me, i guess my question is when does this occur? seems like if the device is giving us impossible values that's a problem with the device not the i/o software, and setting this to 0 could mask that problem. i trust ya but can i have it explained to me why it's preferable to have 0 in the recorded values rather than the value itself, which can be clamped in analysis?

miniscope_io/models/stream.py
```diff
@@ -35,6 +35,16 @@ class ADCScaling(MiniscopeConfig):
         11.3,
         description="Voltage divider factor for the Vin voltage",
     )
+    battery_max_voltage: float = Field(
+        10.0,
+        description="Maximum voltage of the battery."
+        "Scaled battery voltage will be 0 if it is greater than this value",
```
sneakers-the-rat, line 41: maybe use -1 to make it unambiguous from the value actually being zero (which should never happen, but thinking about downstream data analysis)

sneakers-the-rat:
Also I think we are getting to the point where we want to have a separate `ScaledValue` model, having a bunch of fields named by convention isn't great and not very DRY.

like

```python
T = TypeVar("T", bound = Union[int, float])

class ScaledValue(BaseModel, Generic[T]):
    factor: T = 1
    maximum: Optional[T] = None
    minimum: Optional[T] = None

    def scale(self, value: T) -> T:
        return np.clip(value * factor, self.minimum, self.maximum)
```

So then for example

```python
class StreamConfig(BaseModel):
    battery_voltage: ScaledValue = ScaledValue(factor = 10, maximum = 20)
    vin_voltage: ScaledValue = ScaledValue(factor=1000, maximum = 20)
```

where idk we could compute that `ScaledValue`'s `factor` param from other values like the bit depth if we want to or else make a subclass of ScaledValue.

```yaml
battery_voltage:
  factor: 100
  maximum: 10
# ...
```

```python
cfg = StreamConfig()
cfg.battery_voltage.scale(100)
```

mio/models/stream.py
```diff
@@ -82,48 +116,61 @@ class StreamBufferHeaderFormat(BufferHeaderFormat):
...
+        )
+        self._input_voltage_scaling = ScaledValue(
+            scaling_factor=(
+                1
+                / (2**self._adc_scale.bitdepth)
+                * self._adc_scale.ref_voltage
+                * self._adc_scale.vin_div_factor
+            ),
+            maximum=self._adc_scale.vin_max_voltage,
+        )
```
t-sasatani, line 159: [Notes for whenever I can come back] Doing this for every header input feels like a waste and makes the header-filling process inconsistent. So, it is probably better to define a helper class that gets StreamBufferHeader and returns scaled values or something that the scaling factor is set once with the device config.

t-sasatani:
> i guess my question is when does this occur? seems like if the device is giving us impossible values that's a problem with the device not the i/o software, and setting this to 0 could mask that problem.

This mostly happens when the channel or DAQ hardware introduces some bit errors. We're using uint32 to send 8-bit ADC values to simplify the buffer structure, so any flip in the higher bits messes up the metadata plot. I'll change the mask value to `-1`.

PR #88, opened 2024-12-12
Added info about what id is for device config about YAML for CLI

hsemwal:
Added description to device config id in stream.py about id for CLI            "This is currently the id located in the first line of the YAML config file "

----
📚 Documentation preview 📚: https://miniscope-io--88.org.readthedocs.build/en/88/

### DIFF
--- a/mio/cli/stream.py
+++ b/mio/cli/stream.py
@@ -26,7 +26,8 @@ def _common_options(fn: Callable) -> Callable:
         "--device_config",
         required=True,
         help=(
-            "Either a config `id` or a path to device config YAML file for streamDaq "
+            "Either a config `id` or a path to device config YAML file for streamDaq. "
+            "This is currently the id located in the first line of the YAML config file "
             "(see models.stream.StreamDevConfig). If path is relative, treated as "
             "relative to the current directory, and then if no matching file is found, "
             "relative to the user `config_dir` (see `mio config --help`)."
--- a/mio/cli/stream.py
+++ b/mio/cli/stream.py
@@ -27,7 +27,7 @@ def _common_options(fn: Callable) -> Callable:
         required=True,
         help=(
             "Either a config `id` or a path to device config YAML file for streamDaq. "
-            "This is currently the id located in the first line of the YAML config file "
+            "If you aren't using `id` you can ignore them."
             "(see models.stream.StreamDevConfig). If path is relative, treated as "
             "relative to the current directory, and then if no matching file is found, "
             "relative to the user `config_dir` (see `mio config --help`)."


sneakers-the-rat:
Is this not covered by saying the `id` field? Though on load the `id` is moved to the front, it would still work if it wasn't the first line in the file.

sneakers-the-rat:
Or I guess another way of asking, what would "Config `id`" mean so we can correct the ambiguity without overspecifying it by position in the file?

hsemwal:
For someone (like me) who isn't familiar with what different terms are (or how they're getting called), I had no idea. Is ID individual miniscope ID? Is it a type of miniscope (i.e. distinguishing between miniSTIM and Wirefree). Also I was surprised it can be called by the CLI. 

I also don't have a good suggestion about what it could be called in the future. 

The following: _"Either a config `id` or a path to device config YAML file for streamDaq "_ is good for commenting, but just wanted to bring attention to how it is being interpreted on user end.

sneakers-the-rat:
totally ya want to clarify what is unclear to ppl as they are learning what's going on. 

So when you saw "either a config `id` or path to device config YAML file" what did you think it meant, and what cleared it up?

> Is ID individual miniscope ID? 

Since this is CLI docs we can't do what we would normally do which is link to the `Mixin.id` field directly or do the docs, so would something like "`id` field in a config file" be clearer?

Now that both the config cli PR and the "all configs are yaml now" PR are both merged, we could have like a `mio config --list` to show all the config ids that exist so you know what there is to choose from.

 > Also I was surprised it can be called by the CLI.
 
 Not sure what you mean here, what can be called by the cli?

t-sasatani:
I don't think the problem here is ambiguity. It's just that there's no way to find out what `id` is and how to find it from the user's perspective (of course, unless you read all the documents). So it's pretty natural to get into the situation @hsemwal got in.

I might be missing something, but is there any intended interaction for a user to find these IDs? and can we add it to help? I think something like the `mio config --list` you mentioned was mandatory to switch to the config ID thing.

t-sasatani:
It might also be worth considering an ID type field so that it is clear which ID can be injected into each CLI command.

sneakers-the-rat:
> is there any intended interaction for a user to find these IDs?  I think something like the mio config --list

yep! that was the plan. if it's urgent i can do that right now, should only take a few minutes.

> It might also be worth considering an ID type field to clarify which ID can be injected into each CLI command, etc.

that's what the `mio_model` field is for. so we should be able to find all configs with a matching `mio_model` field and list them, makes sense to me to have that be a generic parameter for any command that takes configuration as input, so e.g. you could do like `mio stream capture --list` to see the configs that can be passed to it, or to list by type do `mio config list --model ModelName`. one sec i'll do this after a snack.

t-sasatani:
Sounds great. It won't stop other projects because the direct YAML injection is available. But if you want people to start using the config system, I think it is urgent to have something like this. It'll be better as an *alpha test* of UX if we can merge this before people get used to custom unintended hacks (*i.e.,* modify this file's field every time updating the package.)

t-sasatani:
And we can ask @hsemwal to review whether the designed interaction seems good. We've looked at this too much to have a neutral view.

sneakers-the-rat:
yep! agreed it's an essential part of the design. was doing it in pieces and it's all backwards compatible. if you aren't using ids you can basically ignore them, and i think we'll always support passing a path, so no risk of that going away. The other fields should generate themselves if they aren't there.

To me the design is pretty clear but i can tell there needs to be lots of work on the docs! Configs have ids, you can refer to a config by its id :).

t-sasatani:
Adding this in the help would be the most quick fix. It gets rid of the roadblocks discussed here.
> if you aren't using ids you can basically ignore them.

Yeah, designs are always clear to the people who designed them. I agree it's pretty clear if you read the code, but we probably want to send out reviews to several people for this stuff to learn a user's perspective.

sneakers-the-rat:
Definitely agreed, and yes i was being a little sarcastic with "pretty clear to me" because it is obviously not to other people!

mio/cli/stream.py
```diff
@@ -26,7 +26,8 @@ def _common_options(fn: Callable) -> Callable:
         "--device_config",
         required=True,
         help=(
-            "Either a config `id` or a path to device config YAML file for streamDaq "
+            "Either a config `id` or a path to device config YAML file for streamDaq. "
+            "This is currently the id located in the first line of the YAML config file "
```
t-sasatani, line 30:
```suggestion
            "If you aren't using `id` you can ignore them."
```

t-sasatani (changes requested):
Ok, if @hsemwal agrees with the above update, I think we can merge this. The config interface could be a different PR but totally based on @hsemwal and @sneakers-the-rat 's preference.

mio/cli/stream.py
sneakers-the-rat: I'm still not exactly sure what additional information this gives because i'm not exactly sure what is ambiguous about the current message, i would figure this would be covered by the "either a config `id` or a path" where if i didn't want to use the id field i would use a path, but if this is clearer to someone who isn't me, then i'm fine with it

t-sasatani: Exactly, it does not add anything logic-wise! It just prevents someone from thinking, "What's this ID thing?" before we prepare a feasible way to figure it out (which you are preparing in #93 ).

t-sasatani:
We're good with this, so pls go ahead and do a merge pull request if it looks clearer @hsemwal .

PR #92, opened 2025-01-14
Lowpass mean intensity

MarcelMB:
### Correction for fluctuations of the brightness of the entire field of view

- calculate the mean pixel intensity for every frame
- apply a butter low pass filter to mean intensity over time
creates a figure using the interactive start_frame to end_frame config settings, showing raw data and butter filtered data to visually assess if the filter resembles the data or filter setting need to be adjusted in the config
Screenshot 2025-01-14 at 11 23 15 AM

after pressing 'Esc' script continues running

- scaled images in interactive plot= scaled image by the percentage difference (scaling factor) of raw and butter filtered mean intensity
- this filter is applied to the already spatial mask-filtered data

similar method used in Daniels notebook from v4 Miniscope
[ipynb notebook](https://github.com/Aharoni-Lab/Miniscope-v4/blob/master/Miniscope-v4-Denoising-Notebook/V4_Miniscope_noise_removal.ipynb)
he used it to: 'correct the other source of V4 Miniscope noise. This is a fast, ~+3Hz fluctuation of the brightness'

After closing interactive plot the average and maximum difference between the non-scaled and scaled video is displayed:

Average difference per frame freq_filtered vs scaled_images: 2.247828245162964                                     
Maximum difference across all frames freq_filtered vs scaled_images: 10.018150329589844 

which seems like it has a rather small effect on frame intensity

choose filter settings here
`butter_filter:
  enable: true
  order: 3
  cutoff_frequency: 3
  sampling_rate: 20`

I'm not an expert in if I stick to the general best practice coding of mio, but I included a bunch of logging in case of errors and generally tried to tick to the format of video.py

----
📚 Documentation preview 📚: https://miniscope-io--92.org.readthedocs.build/en/92/

### DIFF
--- a/mio/data/config/process/denoise_example.yml
+++ b/mio/data/config/process/denoise_example.yml
@@ -20,12 +20,12 @@ frequency_masking:
   spatial_LPF_cutoff_radius: 15
   vertical_BEF_cutoff: 2
   horizontal_BEF_cutoff: 0
-  display_mask: true
+  display_mask: false
   output_mask: true
   output_result: true
   output_freq_domain: false
 butter_filter:
-  enable: false
+  enable: true
   order: 3
   cutoff_frequency: 3
   sampling_rate: 20
--- a/mio/data/config/process/denoise_example.yml
+++ b/mio/data/config/process/denoise_example.yml
@@ -2,7 +2,7 @@ id: denoise_example
 mio_model: tests.test_mixins.LoaderModel
 mio_version: 0.6.1.dev16+g6436633.d20241211
 interactive_display:
-  enable: false
+  enable: true
   start_frame: 40
   end_frame: 140
 noise_patch:
@@ -20,10 +20,16 @@ frequency_masking:
   spatial_LPF_cutoff_radius: 15
   vertical_BEF_cutoff: 2
   horizontal_BEF_cutoff: 0
-  display_mask: false
+  display_mask: true
   output_mask: true
   output_result: true
   output_freq_domain: false
+butter_filter:
+  enable: false
+  order: 3
+  cutoff_frequency: 3
+  sampling_rate: 20
+
 end_frame: -1 #-1 means all frames
 output_result: true
 output_dir: user_dir/output
--- a/mio/models/process.py
+++ b/mio/models/process.py
@@ -110,6 +110,27 @@ class FreqencyMaskingConfig(BaseModel):
         description="Whether to output the frequency domain.",
     )
 
+class ButterFilterConfig(BaseModel):
+    """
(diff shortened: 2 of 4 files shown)


t-sasatani:
I sent in @MarcelMB in DM, but I think I will quickly finish getting `feat-preprocess` in main so that this PR's destination can be changed later to `main` to be an independent/focused PR.

sneakers-the-rat:
either way! it's easy enough for me to read as is. we talked a bit earlier and i was going to use this as a demo case for what i was talking about in the other PR, so either you go first or i do on making some structure for these processing routines :)

t-sasatani:
Yeah, as it's not directed to the main, it makes sense to merge instantly, but we don't want the formatting/testing/structuring/discussing/reviewing to accumulate there, so I preferred doing two independent PRs.

That way, the first one will focus on making the base structure, so I was thinking of quickly doing that. I can work on that for a few hours today, but if you have a specific format/preference in mind, I'm happy to let go of it too :-) Either way.

sneakers-the-rat:
i intended to do this today but got caught up on a bunch of other issues that took way longer than i expected. 

Basically all I was going to do here was do something like

```python
class FrameProcessor(ABC):
    @abstractmethod
    def process(self, frame: np.ndarray) -> np.ndarray: ...

    def plot(self, fig: Optional[plt.Figure] = None, ax: Optional[plt.Axes] = None) -> tuple[plt.Figure, plt.Axes]:
        """Optionally plot the results of this processing node"""
        raise 

class ButterFilter(FrameProcessor):
    def __init__(self, config: ButterFilterConfig):
        self.config = config
        # init the rest of the list storing data and stuff
    
    def process(...):
        # move all the processing code here.
        # (well not literally all into one method, probably organize it in some other methods, but expose it in this one method)
        
    def plot(...):
        # move all the plotting stuff here, either taking an existing figure/axes pair or making a new one if not passed. 
        # remove any blocking plot calls so that the calling method can handle the display of plots from multiple processing nodes
```

as we have done elsewhere. i figure that's probably close to what you had in mind, so yno something in that ballpark is what i was thinking but i'm not that picky about it in this case

t-sasatani:
Yeah, something along that. I was wondering a bit about some variations in input/output formats when making everything subclass, so if I don't figure it out a nice way, I'll ask or hand it over (or maybe just Unions).

MarcelMB:
closing this, because it was easier to make a new branch of feature_process right now than to merge feature_process into here and than back. There were a lot of changed to the general pipeline that it was easier to implement this right away of the current commit of feature_process.
so opening a new merge

PR #93, opened 2025-01-15
[cli] Add `mio config list`

sneakers-the-rat:
Re: https://github.com/Aharoni-Lab/mio/pull/88#issuecomment-2591743568

how does this look

Screenshot 2025-01-15 at 3 33 22 AM

### DIFF
--- a/mio/cli/config.py
+++ b/mio/cli/config.py
@@ -2,13 +2,18 @@
 CLI commands for configuration
 """
 
+import re
 from pathlib import Path
 
 import click
 import yaml
+from rich.console import Console
+from rich.table import Table
 
+from mio import CONFIG_DIR
 from mio.models import config as _config
 from mio.models.config import set_user_dir
+from mio.models.mixins import ConfigYAMLMixin
 
 
 @click.group(invoke_without_command=True)
@@ -146,3 +151,41 @@ def user_path() -> None:
     """Location of the current user config"""
     path = list(_config.Config().user_dir.glob("mio_config.*"))[0]
     click.echo(str(path))
+
+
+@config.command("list")
+@click.option(
+    "-v",
+    "--verbose",
+    count=True,
+    help="Display more verbose output\n" "-v: Display full model identifiers and config paths",
+)
+def _list(verbose: int) -> None:
+    """
+    Display the available configs in the user config directory and provided by mio.
+
+    By default, results are truncated for narrow consoles -
+    models within `mio.models` are displayed with only a leading `.`,
+    and paths of builtin configs are shown relative to the package config directory
+    """
+    config_headers = [cfg for cfg in ConfigYAMLMixin.iter_configs()]
+    config_headers = sorted(config_headers, key=lambda cfg: (cfg["mio_model"], cfg["id"]))
+
+    table = Table(title="mio configs")
+    table.add_column("id", style="yellow", no_wrap=True)
+    table.add_column("mio_model")
+    table.add_column("path")
+
+    for header in config_headers:
+        table.add_row(
+            header["id"],
+            re.sub(r"^mio.models", "", header["mio_model"]) if verbose < 1 else header["mio_model"],
+            (
+                str(header["path"].relative_to(CONFIG_DIR))
+                if verbose < 1 and CONFIG_DIR in header["path"].parents
+                else str(header["path"])
+            ),
+        )
+
+    console = Console()
+    console.print(table)
--- a/mio/models/mixins.py
+++ b/mio/models/mixins.py
@@ -5,15 +5,33 @@ to use composition for functionality and inheritance for semantics.
 
 import re
 import shutil
+import sys
 from importlib.metadata import version
 from itertools import chain
 from pathlib import Path
-from typing import Any, ClassVar, List, Literal, Optional, Type, TypeVar, Union, overload
+from typing import (
+    Any,
+    ClassVar,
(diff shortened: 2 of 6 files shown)


t-sasatani:
Nice!

sneakers-the-rat:
i'll do `mio config show {id}` tomorrow so you can then take a look at them

and then maybe we want like `mio config path {id}` to output the path so you can do something like `subl "$(mio config path my-config)"` to open it in a text editor

t-sasatani:
I think it's already satisfactory for addressing #88 , though I would want to hear from others how it feels to navigate through this. And yeah, the intended interaction for changing configs was another part I wasn't sure about, so an interface for grabbing paths would be excellent.

MarcelMB:
Thanks, Jonny for creating this.

Just for my understanding: Why is there a stream-buffer-header ID and a wireless-200px ID.

So it's not like streaming with a buffer-header file with the wireless MS is going to be one ID? For example, each pixel size on the image sensor will have a different ID? and will be combined with the wireless-XXpx file?

So I think what I am saying is I am not sure if it would be intuitive to know which configuration to use.

sneakers-the-rat:
So see the second column, which says what kind of model it is. currently that's just the name of the class, but in the future once we have done more work to link e.g. each of the other cli commands to a given config model (the `mio stream capture` command takes a `StreamDevConfig` object) then we can refine the interface in a few ways: 
- column for showing which commands each config can be used with, e.g. like `mio config list --with-commands` or something.
- parameter for showing 
- `--list` parameter for each of the other commands that lists configs that have a matching model type. e.g. `mio stream capture --list` would show you a table of just the stream dev configs you have.
- ... and so on, since we are linking config, code, metadata then we can refine the interface using all that information.

this is just the first draft of a "list all configs" command.  so there are tons of other things we will want here too! this is an example of making small iterable PRs - initial functionality with clear points where we can expand in the future

the idea behind the design of this `id`, `mio_model`, `version` header thing coupled to the pydantic models is to have a uniform interface across API, cli, configs, and within the package. Since this is an SDK and we expect a lot of different pieces to need to work together in development, and also we expect users to need to have a durable, versionable, publishable description of their experiments, this package needs a lot of configuration. On two extremes, we can either have gigantic config files that jam everything into a humongous model (this is sort of how kilosort is/was designed, a single huge configuration blob gets passed around the whole program), or have to constantly juggle a bunch of different configs and variables. 

I agree it's sort of annoying to need to have both the `StreamDevConfig` and the `BufferHeaderFormat` (usually the header format is just supplied by default), but this `id` and `model` system embedded in the configs lets us resolve that. They are actually coupled right? like the format of header data should be properly considered part of the device configuration. Say we had a field `header_format` in `StreamDevConfig` that needs a `BufferHeaderFormat`. We can specify that by reference everywhere in the same way without needing to juggle paths.

<table><thead><tr><th>mode</th><th>`id`/`mio_model`</th><th>previously</th></tr></thead><tbody><tr><td>config</td><td>

```yaml
header_format: wireless-header-format
```

</td><td>

```yaml
header_format:
  linked_list: 0
  frame_num: 1
  # ... etc, literally in the config file
```

or

```yaml
header_format: ./buffer_header_format.yaml

```

</td></tr><tr><td>cli</td><td>

```bash
# use the format specified in the config
mio stream capture wireless-200px

# override values in the config
mio stream capture wireless-200px \ 
  --with header_format=wireless-header-format
```

</td><td>

```bash
# to use the default configs from a non-git, non-development install

mio stream capture ../venv/lib/site-packages/mio/config/wireless_200px.yaml \
  --header-format ../venv/lib/site-packages/mio/config/wireless_200px.yaml

```
</td></tr>
<tr><td>code</td><td>

```python
# load the config
StreamDevConfig.from_id('wireless-200px')

# to specify the header config
StreamDevConfig(header_format='wireless-header-format')
```

</td><td>

```python
from mio import CONFIG_DIR

# load the config
StreamDevConfig.from_yaml(CONFIG_DIR / "wireless_200px.yaml")

# specify header config
StreamDevConfig(
    header_format=BufferHeaderFormat.from_yaml(
        CONFIG_DIR / "wireless_header_format.yaml"
    )
)

```

</td></tr></tbody></table>

So that's what we're working towards, where we can break apart, recombine, etc. different configs together both provided by the package, within an experiment, across a user's machine, or included in some downstream package that bundles us. 

Hopefully that help explains why we would be seeing all configs and why that's a good thing - we *want* to be able to have a bunch of different types of configs for different models: those are all the configs and their ids you can use *anywhere*, in different cli commands, within config files, etc. And yes we will certainly continue to refine this interface and make use of that information to be able to filter and sort and show what you need to know, this isn't the final PR of the CLI. We will need to draw the line somewhere on this package to keep it functional for development and use in downstream packages, but doesn't try to bundle a ton of GUI and other stuff that we want to purposely keep separate to maintain modularity, but the CLI should be good.

t-sasatani (approved):

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "platformdirs>=4.3.6",
     'typing-extensions>=4.10.0; python_version<"3.13"',
     "scikit-video>=1.1.11",
+    "scipy>=1.13.0",
```
sneakers-the-rat, line 27: made a lockfile that splits out <3.10 and >=3.10, noticed that we are using scipy within the package but don't have it as a direct dependency

PR #95, opened 2025-01-17
[ci] Run tests against any PR

sneakers-the-rat:
Since we're doing stacked up PRs, we should be running tests against each of them to avoid errors piling up

----
📚 Documentation preview 📚: https://miniscope-io--95.org.readthedocs.build/en/95/


sneakers-the-rat:
gonna self-approve this since it's tiny, hopefully uncontroversial, doesn't affect the operation of the package, and is easily reversible

Issue #42, opened 2024-08-28
Discussion: prefix design and matching


sneakers-the-rat:
i think we've talked about this in a few issues and in the slack but i don't think we have a single issue for this - chattin about what we want for recognizing boundaries between frames.

detecting frames currently relies on a match between a prefix value:
- miniscope-io config: https://github.com/Aharoni-Lab/miniscope-io/blob/302c6909e789431c406aeea00edc24dc238670de/miniscope_io/data/config/WLMS_v02_200px.yml#L12
- samd config: https://github.com/Aharoni-Lab/Miniscope-SAMD-Framework/blob/b85a4867073166a255dbcc6355e373324eda264e/include/MS_config.h#L70
- detection: https://github.com/Aharoni-Lab/miniscope-io/blob/302c6909e789431c406aeea00edc24dc238670de/miniscope_io/stream_daq.py#L325

Error characterization:
The current problem i know of is that there is some known rate of bit corruption that makes the prefix not a perfect match. but plz list any others. Knowing exactly what kind of corruption we need to be robust to would probably go a a long way - eg. if it's just bit flips we could do a convolution, but if there are insertions/deletions it wouldn't work so good.

constraints:
- small: want to not spend all day on the preamble
- specific: no false positives in the data
- sensitive: tolerant to mild corruption
- ?

Some ideas for strategies:
- Fuzzy matching: sum of an XOR above some value? pull in a levenshtein?
- Range guessing: currently we just search for any match in the string, but we could allow for even more fuzziness and make the search faster by seeding searches in the place we expect the next preamble to be in the bitstream (assuming 'packets' are uniform length?)
- Repeating preamble: the preamble is currently `0x12345678`, is there some preamble that is empirically worse or better than another, or is every preamble equally likely in the data? i also wonder if we can use a repeating preamble, eg. `10101...` so that we can scale it up and down depending on the quality of the signal, as well as set a threshold in miniscope-io? so like "treat any sequence that ends with 4 copies of the preamble as a buffer, if we detect too many, increase that threshold, if we detect too few, decrease it"

there is also probably some encoding magic like that manchester encoding dummy value thing that @t-sasatani just pulled in the SAMD framework that i am not aware of, so plz fill in the gaps here and i can edit them into this top comment until we come up with a plan

also @MarcelMB @phildong

t-sasatani:
[...]

PR #97, opened 2025-01-29
extensive test for gradient detection

MarcelMB:
added a much larger and more diverse .avi file (wireless_corrupted_extended) with 872 frames manually sorted by me and taken from wireless in vivo imaging to test if we can find broken buffers successfully

right now only for gradient method because the mean method had some issues as described issue #42

the cool thing about if the test failed that it shows how many frames were missed and in which category frames were missed, the different types of error categories are explained in the added doc file (preprocessing video files).
This allows for a better insight into what to improve for broken buffer detection methods and compares between them

Screenshot 2025-01-29 at 3 30 54 PM

also added docs in API: preprocessing video files (needs to be further updated but is a start)

Screenshot 2025-01-29 at 3 31 13 PM
Screenshot 2025-01-29 at 3 31 23 PM

first test I ever wrote. Which is strongly based on Jonnys test they wrote earlier. But let me know if this needs changed/improvements before merging

Right now this test fails because the broken buffer method cant detect black-out buffers and misses some broken buffers where only a few rows of pixels are broken or less than one row of pixels

needs an update on the broken buffer detection in a different PR (and combining different methods as we discussed above)

----
📚 Documentation preview 📚: https://miniscope-io--97.org.readthedocs.build/en/97/


t-sasatani:
Seems cool! Before getting into details, I wondered whether this function should be a CI test like this PR or a CLI option like `mio denoise -c some_config --benchmark` or `--verbose`. For example, the gradient not capturing blackout buffers or small stuff below the threshold is an expected behavior. So, if we're doing this as tests, we should expect them not to be captured. Maybe @sneakers-the-rat knows a good way to organize this.

sneakers-the-rat:
making a proper user-exposed benchmarking function wouldn't be a bad idea. for that we'd want a bit of work on properly modeling the different kinds of errors, making a runner for it that encapsulates the preprocessing pipeline, etc. If we had it for end users then the tests for the preprocessing methods themselves could just use that too, so benchmarking cli utility and CI tests are complementary :)

docs/index.md
```diff
@@ -31,6 +31,7 @@ api/exceptions
 api/plots/index
 api/utils
 api/stream_daq
+api/preprocessing
```
sneakers-the-rat, line 34: I think the docs need a decent amount of reorganizing, so this is fine for now, but something like this should probably go in a "user guide" section rather than API docs. but since we don't have a proper user guide yet we can handle that when we do.

tests/data/wireless_corrupted_extended.avi
sneakers-the-rat, line 1: A 50MB test data file is pretty dang big. The purpose of the smaller test files is that they contain exactly what we want to test and no more. Since these are just test data, if we want to keep this, it would be great to compress this video: this video is 1MB when compressed with default x264 settings.

tests/data/wireless_corrupted_extended.yaml
```diff
@@ -0,0 +1,95 @@
+frames: # frame indexing in cv2 starts from 0
+  check_pattern:
+    less_than_one_row:
```
sneakers-the-rat, line 3: I'm not going to be a stickler here, but these categories are a *little* squishy to me. What we're trying to capture is (a) the type of corruption (noise vs. blacked out) and (b) the size of that corruption. Ideally (b) would be quantitative, like a count of the pixels or the count of affected rows like this:

```yaml
frames:
  check_pattern:
    - frame: 0
      rows: 10
    - frame: 10
      rows: 14         
```

and that way we can get a quantitative threshold on what we are detecting vs. not detecting.

as-is i'm not *really* sure what the difference between the different subcategories is, but since it's not something in the package itself, but just in the tests, it's not that big of a deal to me. if we want to make this into a user-exposed benchmarking thing then i think we'd want to change it.

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat, line 1: I think this should go in the `test_frame_helper.py` module - we want the test modules to reflect the structure of the package, so someone knows to go looking for the tests for `mio.process.frame_helper` in `tests/test_process/test_frame_helper.py`

tests/test_process/test_extended_frame_helper.py
```diff
@@ -0,0 +1,85 @@
...
+from ..conftest import DATA_DIR
+
+
+@pytest.mark.parametrize(
+    "noise_detection_method",
+    [
+        "gradient",  # Only testing the "gradient" method
+    ],
+)
+def test_noise_detection_contrast_with_categories(noise_detection_method):
```
sneakers-the-rat, line 18: ya ya as above, i think this should either replace the existing test function, or else we should add video/ground truth as a parameter and run this test over several different videos. no reason to duplicate code here

tests/test_process/test_extended_frame_helper.py
```diff
@@ -0,0 +1,85 @@
...
+        is_noisy, _ = detector.detect_frame_with_noisy_buffer(
+            current_frame=frame, previous_frame=None, config=config
+        )
+
+        detected_frames.append(is_noisy)
+
+    detected_frame_indices = np.where(detected_frames)[0]
+
+    # Analyze missed frames by category
+    missed_counts = {}
```
sneakers-the-rat, line 58: ok so what we want here - we want the tests to tell us when different categories of errors are missed. The means by which we make the tests tell us that something in particular is broken is to have a unit test that is specific to that thing, and to have that test fail. So for each category of corruption and each type of noise detector we want to have a "separate" test - in quotes because "separate" also includes test parameterizations. So I think that we should make the different categories (noise vs. blackout) a parameter so that we can say "noise detector x is failing for noise category y" easily.

I don't really know what we get from splitting the ground truth out by the size of the corruption - the ground truth should be the ground truth, either we detect it or we don't, and missing even the smallest patch in the ground truth should be a test failure the same as if we missed the biggest patch. both are regressions (that's what this test is protecting us from, that in the future we don't accidentally break the noise detectors somehow). But if this is useful information to y'all we can definitely keep it in

tests/test_process/test_extended_frame_helper.py
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+    # Print or assert results for debugging
+    for category, subcategories in missed_counts.items():
+        for subcategory, stats in subcategories.items():
+            print(
+                f"{category} -> {subcategory}: Missed {stats['missed_count']} "
+                f"out of {stats['total_expected']} frames."
+            )
+            print(f"Missed Frames: {stats['missed_frames']}")
```
sneakers-the-rat, line 78: no print statements in tests, we do this with assertions :). these shouldn't even show up for you if you are running tests via pytest
```suggestion
```

tests/test_process/test_extended_frame_helper.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Example assertion: Ensure no missed frames for critical categories
+    assert all(
+        stats["missed_count"] == 0
+        for category in missed_counts.values()
+        for stats in category.values()
+    ), "Some frames were missed in one or more categories!"
```
sneakers-the-rat, line 85: here's how I would write this (we would need to add `from pprint import pformat` at the top).

We are missing false-positives (`extra_frames`) that doing an `array_equal` check gives us in the current test, and as-written we would not get the detail in the test output that this PR is intended to introduce - the error message that we actually see would just be "some frames were missed, {int} != 0" or something like that. This would show all the false positives/negatives, if any. 

```suggestion
    missed = []
    
    for category, subcategories in expected["frames"].items():
        for subcategory, expected_frames in subcategories.items():
            expected_set = set(expected_frames)
            detected_set = set(detected_frame_indices)
            missed_frames = expected_set - detected_set  # Frames in expected but not detected
            extra_frames = detected_set - expected_set
            if missed_frames or extra_frames:
                missed.append({
                    "missed": missed_frames,
                    "extra": extra_frames,
                    "category": category,
                    "subcategory": subcategory
                })

    assert missed == [], f"Missed or detected extra frames with method {noise_detection_method}: {pformat(missed)}"
```

sneakers-the-rat (commented):
Love this. 

Only thing I think is 'must fix' is adding false-positive detection back in.

I would relatively strongly prefer to not add a 50MB test file here - it's already in the repo history and we'd need to use `git filter-repo` to actually recover the space, but if we can compress that video that would be great (i can do that in a sec).

Other suggestions are to not make a separate test module for this, and either combine it (make the ground truth annotations match this new format, and parameterize the test over video samples) or replace the existing test, but otherwise looks good to me.

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat: I'll do this rn

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat: disregard this, need a different way to detect false positives lol, one sec doing this now

sneakers-the-rat: Done in https://github.com/Aharoni-Lab/mio/pull/97/commits/3fdb6e44840b3c22e6274ec0e726fe4e6eb062bb

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat: Done in https://github.com/Aharoni-Lab/mio/pull/97/commits/3fdb6e44840b3c22e6274ec0e726fe4e6eb062bb

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat: Done in https://github.com/Aharoni-Lab/mio/pull/97/commits/3fdb6e44840b3c22e6274ec0e726fe4e6eb062bb

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat: done in https://github.com/Aharoni-Lab/mio/pull/97/commits/3fdb6e44840b3c22e6274ec0e726fe4e6eb062bb

tests/test_process/test_extended_frame_helper.py
sneakers-the-rat: done in https://github.com/Aharoni-Lab/mio/pull/97/commits/3fdb6e44840b3c22e6274ec0e726fe4e6eb062bb

sneakers-the-rat:
Alright i merged the new test with the existing test - it seems like each of these noise detection methods is designed to test a different kind of noise, so i made it so we test a given noise detection method along with a category of frame corruption (rather than testing all methods * all categories). Also did the things i mentioned in the review (expand those comments to see rationale, unresolve and comment there if there is discussion needed).

We've still got a few problems:
- The `mean_error` method is detecting basically every frame in that extended video as noisy. This seems to stem from the fact that we don't store the previous frame to compare the next against when it's determined to be noisy. The first frame in this video is noisy, and so we never seem to get a stable baseline to compare against. This is a runaway problem, where once we get more than few frames away from the last good frame, it's unlikely that we'll detect a frame as non-noisy again because the image will actually be different by then. I think that mean error method might not really be what we want , since it seems like the gradient method is better at detecting the speckled noise pattern anyway, and the other category of error from having big blanked out sections of the video should be done with a more specific detection method (should be easy enough... like `np.sum(frame < np.median(frame)/5, axis=0) > threshold` )
- the gradient method is missing several of the `less_than_one_row` frames, which is sort of to be expected since they are the smallest ones. that might be just a threshold adjustment, but it also might mean we need to make that detection method row-wise rather than chunk-wise (which would be more performant anyways, probably).

t-sasatani (approved):
I agree we can drop mean error if gradient works better. Requiring multiple frames isn't ideal anyways.

I initially thought working on the serialized frame and chunking with the actual buffer boundary in mind would be better, but it might not matter for errors small as a single line. (and yeah numpy is probably faster just doing stuff in rows.)

> we don't store the previous frame to compare the next against when it's determined to be noisy.

But this might be a bug introduced in recent PRs or just a threshold problem? The unit test with a small video clip passed so it should have been storing previous frames or otherwise it would have been mostly false positives.

I'm reviewing on my phone so haven't gone through details but I think it looks good if we're just skipping mean error part (and this doesn't go to main branch anyways)

sneakers-the-rat:
ha ok check out https://github.com/Aharoni-Lab/mio/pull/97/commits/77a03a5193cb23b1e0a2ddd790f1eeca0039c2ee

using just the row-wise gradients actually works way better, we don't even need to iterate over chunks. This makes sense because the glitches accumulate along a row, but not columns, so all the columnwise gradients do are wash out the signal if it's present.

it's so sensitive and specific that it didn't detect any non-noise frames, detected all the noisy frames including frame 70 which just has part of a single row corrupted (below), AND it found a noisy frame that is in fact noisy but wasn't in the ground truth (582).

Screenshot 2025-01-29 at 7 45 36 PM

Still have the problem with the mean error method detecting every frame as noise. We can xfail that in this PR and pick it up in the main PR since this PR wasn't supposed to be about the `mean_error` method.

t-sasatani:
> glitches accumulate along a row, but not columns,

So this part might be a minor misunderstanding. It should accumulate in the serialized form of the image, with borders at the end of the buffer. However, just doing array manipulation is much much better for simplicity so if row-wise work let's do row-wise.

sneakers-the-rat:
> I agree we can drop mean error if gradient works better. Requiring multiple frames isn't ideal anyways.

OK i'll mark it as xfail and merge into `feat-preprocess`, we can decide what we want to do before we merge that back to main.

> But this might be a bug introduced in recent PRs or just a threshold problem?

It looks like this is how it has been working: https://github.com/Aharoni-Lab/mio/blob/9c04b7d27e465a388cab44ec612337c00f1c2690/mio/process/video.py#L168-L179

where `self.previous_frame` is only assigned to if the frame is not `broken`. It should only be a problem if the first frame is detected as noisy (as this video is), or if there are long runs of noisy frames, so we probably haven't encountered it yet bc the problem is masked by it mostly working fine.

> So this part might be a minor misunderstanding. It should accumulate in the serialized form of the image, with borders at the end of the buffer.

Right, yes. i meant that in the reassembled frames it appears to be row-wise because the pixel-wise stream is row-major

t-sasatani:
ok thanks failing with long runs of broken frames make total sense.

sneakers-the-rat:
OK I switched the uncompressed video out with a compressed version - i wanted to make it smaller, preserve fine enough detail that it would stay useful for testing noise, and preserve editing ability in FIJI, which can only handle `baseline` x264 videos, so it's sort of a rough compromise. but yeah we do want to keep this package under like ~100MB to clone or so or else it gets annoying for ppl to work with.

git stores all files that are committed, even if they're removed, so this is actually a net addition to the repo size for now, but once the parent branch is merged i'll go back and remove it

PR #98, opened 2025-01-30
changed gradient method to start with 1st frame

MarcelMB:
changed detect_frame_with_noisy_buffer in video.py

If the method is "gradient", the detect_frame_with_noisy_buffer method is called immediately on the first frame, without needing a previous frame
1st frame in a video was previously skipped, which is fine for the mean method since it needs a previous frame

was important since I also implemented a test where the 1st frame has a broken buffer and this could not be detected earlier

gradient method first frame processed:
Screenshot 2025-01-29 at 4 21 39 PM

mean method first frame skipped:
Screenshot 2025-01-29 at 4 21 25 PM

----
📚 Documentation preview 📚: https://miniscope-io--98.org.readthedocs.build/en/98/

### DIFF
--- a/mio/process/video.py
+++ b/mio/process/video.py
@@ -160,12 +160,28 @@ class NoisePatchProcessor(BaseVideoProcessor):
         if input_frame is None:
             return None
 
-        if self.noise_patch_config.enable and self.previous_frame is not None:
-            broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
+        if self.noise_patch_config.enable:
+            if self.noise_patch_config.method == "gradient":
+            # For the gradient method, analyze the first frame immediately
+                broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
                 input_frame,
-                self.previous_frame if self.noise_patch_config.method == "mean_error" else None,
+                None,  # No previous frame needed for gradient method
                 self.noise_patch_config,
             )
+            else:
+                # For the mean method, wait until the second frame to start analysis
+                if self.previous_frame is not None:
+                    broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
+                        input_frame,
+                        self.previous_frame,
+                        self.noise_patch_config,
+                )
+                else:
+                    # If it's the first frame and the method is mean_error, just store the frame
+                    self.previous_frame = input_frame
+                    self.append_output_frame(input_frame)
+                    return input_frame
+                
 
             # Handle noisy frames
             if not broken:
@@ -181,10 +197,8 @@ class NoisePatchProcessor(BaseVideoProcessor):
                 self.dropped_frame_indices.append(index)
             return None
 
-        elif self.noise_patch_config.enable and self.previous_frame is None:
-            self.previous_frame = input_frame
-            self.append_output_frame(input_frame)
-            return input_frame
+        self.append_output_frame(input_frame)
+        return input_frame
 
     @property
     def noise_patch_named_video(self) -> NamedVideo:
--- a/mio/process/video.py
+++ b/mio/process/video.py
@@ -162,12 +162,12 @@ class NoisePatchProcessor(BaseVideoProcessor):
 
         if self.noise_patch_config.enable:
             if self.noise_patch_config.method == "gradient":
-            # For the gradient method, analyze the first frame immediately
+                # For the gradient method, analyze the first frame immediately
                 broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
-                input_frame,
-                None,  # No previous frame needed for gradient method
-                self.noise_patch_config,
-            )
+                    input_frame,
+                    None,  # No previous frame needed for gradient method
+                    self.noise_patch_config,
+                )
             else:
                 # For the mean method, wait until the second frame to start analysis
                 if self.previous_frame is not None:
@@ -175,13 +175,12 @@ class NoisePatchProcessor(BaseVideoProcessor):
                         input_frame,
                         self.previous_frame,
                         self.noise_patch_config,
-                )
+                    )
                 else:
                     # If it's the first frame and the method is mean_error, just store the frame
                     self.previous_frame = input_frame
                     self.append_output_frame(input_frame)
                     return input_frame
-                
 
             # Handle noisy frames
             if not broken:


mio/process/video.py
```diff
@@ -156,13 +156,28 @@ def process_frame(self, input_frame: np.ndarray) -> Optional[np.ndarray]:
...
+        if self.noise_patch_config.enable:
+            if self.noise_patch_config.method == "gradient":
+            # For the gradient method, analyze the first frame immediately
+                broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
                 input_frame,
-                self.previous_frame if self.noise_patch_config.method == "mean_error" else None,
+                None,  # No previous frame needed for gradient method
                 self.noise_patch_config,
             )
+            else:
```
t-sasatani, line 167:
```suggestion
            if self.noise_patch_config.method == "mean_error":
```
Probably better to make it explicit here.

mio/process/video.py
```diff
@@ -156,13 +156,28 @@ def process_frame(self, input_frame: np.ndarray) -> Optional[np.ndarray]:
...
+                        input_frame,
+                        self.previous_frame,
+                        self.noise_patch_config,
+                )
+                else:
+                    # If it's the first frame and the method is mean_error, just store the frame
+                    self.previous_frame = input_frame
+                    self.append_output_frame(input_frame)
+                    return input_frame
+
```
t-sasatani: Raise not implemented error if it's not any of the above. We might need to import `NotImplementedError`.
```suggestion
            else:
                raise NotImplementedError(f"Noise detection method {self.noise_patch_config.method} is not implemented.")

```

mio/process/video.py
```diff
@@ -178,11 +193,8 @@ def process_frame(self, input_frame: np.ndarray) -> Optional[np.ndarray]:
                 self.dropped_frame_indices.append(index)
             return None
 
-        # Initialize previous frame if it's the first frame
-        if self.noise_patch_config.enable and self.previous_frame is None:
-            self.previous_frame = input_frame
-            self.append_output_frame(input_frame)
-            return input_frame
+        self.append_output_frame(input_frame)
+        return input_frame
```
t-sasatani, line 196: Yeah I agree it's better to take care of this part in the method `if`.

t-sasatani (approved):
I think this is good and can go ahead with some minor changes. My suggestions are minor ones just for adding more detection methods in the future.

mio/process/video.py
```diff
@@ -156,13 +156,27 @@ def process_frame(self, input_frame: np.ndarray) -> Optional[np.ndarray]:
...
+                    broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
+                        input_frame,
+                        self.previous_frame,
+                        self.noise_patch_config,
+                    )
+                else:
+                    # If it's the first frame and the method is mean_error, just store the frame
+                    self.previous_frame = input_frame
+                    self.append_output_frame(input_frame)
+                    return input_frame
```
t-sasatani, line 180: Noticed it got outdated, so just updating here:
```suggestion
            else:
                raise NotImplementedError(f"Noise detection method {self.noise_patch_config.method} is not implemented.")
```

mio/process/video.py
sneakers-the-rat: probably with an `elif`

mio/process/video.py
```diff
@@ -156,13 +156,27 @@ def process_frame(self, input_frame: np.ndarray) -> Optional[np.ndarray]:
...
             return None
 
-        if self.noise_patch_config.enable and self.previous_frame is not None:
-            # Call the unified detection method
-            broken, noise_patch = self.noise_detect_helper.detect_frame_with_noisy_buffer(
-                input_frame,
-                self.previous_frame if self.noise_patch_config.method == "mean_error" else None,
-                self.noise_patch_config,
-            )
+        if self.noise_patch_config.enable:
```
sneakers-the-rat, line 159: to avoid having this entire method gaining an extra layer of indentation and and more layers of `if/else` switches, this should probably be 

```python
if not self.noise_patch_config.enable:
    return input_frame
```

sneakers-the-rat (approved):
seems fine to me :)

sneakers-the-rat:
just noticing that we're saving copies of all the frames that go through this class. we'll want to change that in the future bc it's an unnecessary use of memory, but fine for now

MarcelMB:
Thanks for the fast response.
I will address the minor changes before merge.

Not sure if this the correct way but I just added another commit here after I had opened the PR. Sorry about that.

I added a lower threshold for the gradient method to also detect successfully the black broken buffer that are 'obvsiously' low in contrast as opposed to the check pattern ones that are high in contrast. SO now it is detecting also the black broken buffer frames. Its just 2-3 lines added to the code in the most recent commit.

mio/process/video.py
t-sasatani: oh yes good catch I updated it

t-sasatani: Oh yes good catch I updated it

t-sasatani:
Adding this lower threshold makes perfect sense. It might be better to split lower and higher thresholding in the future because the lower thing we want to detect should also be zero, and asserting that could decrease false positives, but this can be something later.

sneakers-the-rat:
> Not sure if this the correct way but I just added another commit here after I had opened the PR. Sorry about that.

for the record this is not only totally fine, but expected :).

> I added a lower threshold for the gradient method to also detect successfully the black broken buffer that are 'obvsiously' low in contrast as opposed to the check pattern ones that are high in contrast

I think that we should probably not try and overload this detection method and make detecting blacked out sections a different detection method - recall that this is on the changes between pixels, not pixel value, and it's valid for a frame to have very little/no change between pixels. A separate method that explicitly looks for large blocks of all-zeros would be cleaner and more specific imo

sneakers-the-rat:
Doing this reversion/deconflicting while Marcel is sitting here next to me, so i'm not just going rogue and not consulting with him lol

t-sasatani:
Was this PR closed intentionally or accidentally?

MarcelMB:
We did the revision with Jonny in person and merged it already. If I am not wrong it should be included in the feature_process branch. So that's why I closed it.

t-sasatani:
Oh thanks overlooked this. Ill make the sub branch point to preprocess branch too then.

sneakers-the-rat:
oh wait now i am not sure if i finished that and merged it

t-sasatani:
Oh, right, seems like only the other way around is done. I can also handle this while looking into the black padding tmr if no one else does it.

t-sasatani:
It doesn't make sense to handle this and #99 separately at this point, so I'll re-close this.

PR #99, opened 2025-01-31
Combined detector / black area detector / bugfix process frame index

MarcelMB:
The `_detect_black_pixels` looks if consecutive pixels in a row are black

` black_pixel_consecutive_threshold: 5 # If 5 consecutive pixels in a row are black, discard the frame
  black_pixel_value_threshold: 16 # Any pixel value ≤ 16 is considered "black"`
  
  those two parameters in denoise_examle.yaml decides how many consecutive pixels need to be black and what is defined as black (black pixels don't have a value of zero but when I checked was like 16)
  
 updated `detect_frame_with_noisy_buffer` to work with two methods simultaneously

added `test_frame_helper.py` to the test to also test the gradient method on blacked_out subcategory frames (since the gradient method now also incorporates finding black frames)
test all pass (except the xfail that jonny marked for obvious reasons)

curious about if this makes sense to you guys or if there are any major concerns!

example image in debug mode

Screenshot 2025-01-31 at 2 49 10 PM

detects error in this image:
Screenshot 2025-01-31 at 2 57 00 PM

----
📚 Documentation preview 📚: https://miniscope-io--99.org.readthedocs.build/en/99/

### DIFF
--- a/docs/api/preprocessing_guide.md
+++ b/docs/api/preprocessing_guide.md
@@ -1,4 +1,4 @@
-# preprocessing video files
+# Preprocessing videos
 
 
 ## Filtering of frames with broken buffers (blocks of a frame) 
--- a/docs/api/process/frame_helper.md
+++ b/docs/api/process/frame_helper.md
@@ -1,0 +1,7 @@
+# Frame helpers
+
+```{eval-rst}
+.. automodule:: mio.process.frame_helper
+    :members:
+    :private-members:
+```
--- a/docs/api/process/index.md
+++ b/docs/api/process/index.md
@@ -1,0 +1,9 @@
+# preprocess
+
+```{toctree}
+:caption: Preprocess
+
+video
+frame_helper
+zstack_helper
+```
--- a/docs/api/process/video.md
+++ b/docs/api/process/video.md
@@ -1,0 +1,7 @@
+# Video preprocessor
+
+```{eval-rst}
+.. automodule:: mio.process.video
+    :members:
+    :private-members:
+```
--- a/docs/api/process/zstack_helper.md
+++ b/docs/api/process/zstack_helper.md
@@ -1,0 +1,7 @@
+# z-stack helpers
+
+```{eval-rst}
+.. automodule:: mio.process.zstack_helper
+    :members:
+    :private-members:
+```
--- a/docs/index.md
+++ b/docs/index.md
@@ -9,6 +9,7 @@ Generic I/O interfaces for miniscopes :)
 guide/installation
 guide/config
 cli/index
+api/preprocessing_guide
 ```
 
 ```{toctree}
@@ -31,9 +32,9 @@ api/exceptions
 api/plots/index
 api/utils
 api/stream_daq
-api/preprocessing
 api/bit_operation
 api/vendor/index
+api/process/index
 ```
 
 ```{toctree}
--- a/mio/data/config/process/denoise_example.yml
+++ b/mio/data/config/process/denoise_example.yml
@@ -3,6 +3,14 @@ mio_model: mio.models.process.DenoiseConfig
 mio_version: 0.6.1
 noise_patch:
   enable: true
+  method: gradient #gradient or mean_error or black_pixels
+  additional_methods:
+    - black_pixels
+  threshold: 20 #for first method
+  black_pixel_consecutive_threshold: 5 # If 5 consecutive pixels in a row are black, discard the frame
+  black_pixel_value_threshold: 16 # Any pixel value ≤ 16 is considered "black"
+  buffer_size: 5032
+  buffer_split: 8
   method: gradient # gradient or mean_error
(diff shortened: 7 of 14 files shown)


sneakers-the-rat:
sorry i was being unclear when i said separate "method" - i meant method there as being like "means of detecting noise" not "python method." lets split each of these into a different class with a shared calling convention, bc we want to be able to apply them together (do black pixel detection, gradient detection, and other preprocessing steps) and make it clear which config values correspond to each of those different strategies (we already had one almost-bug from reusing the threshold param) :)

i'll handle this next week

sneakers-the-rat:
question for y'all - is there any time where we would expect there to be black areas, like daniel was mentioning that sometimes there might be an all-black row or column on the sides of the image? I think that consecutive black pixels (as in literally value == 0, not just a low value) is the right test, just checking if there are any edge cases where we might need to e.g. check for two consecutive rows, or check that the black pixels aren't on the borders

t-sasatani:
The (problematic) black regions here should occur because of the black padding in the `streamDaq`. So:
- I think we should start by finding pixels that are strictly 0. Except for the padded black zones, I can't think of any other dark regions we should directly eliminate.
- Finding pixels with intensity under 16 could also eliminate dark areas, depending on the lighting conditions (tho this is unlikely to happen with our setup with low contrast).
- The black rows/columns are something we wish we had and are not present in the image now.
- We should eventually move this into the `streamDaq` or use metadata from that, though I prefer doing it here for now to simplify stuff.

MarcelMB:
> sorry i was being unclear when i said separate "method" - i meant method there as being like "means of detecting noise" not "python method." lets split each of these into a different class with a shared calling convention, bc we want to be able to apply them together (do black pixel detection, gradient detection, and other preprocessing steps)

- that's possible now. In this PR you can run gradient (so check pattern noise) and black detection simultaneously. and they also use different config values not a shared threshold 

>  just checking if there are any edge cases where we might need to e.g. check for two consecutive rows, or check that the black pixels aren't on the borders

- it make sense to be that we don't just check for 5 consecutive black pixels in one row and mark it as black. I think we would not except any black pixels in the real data at least not below a threshold of like 20. But we can change the criteria to say 2 or 3 rows should have black  pixels to me a bit more liberal

- Right now the minimum pixel is always 16 (no 0 pixel value in the data)


> Unfortunately, the always black rows/columns are something we wish we had and are not present in the image now.

- for what would we want to use this? just curious. 

@Jonny how would you structure adding this black method differently? You can go ahead and add commits here and I can take a look.

MarcelMB:
would like to merge this soon to `feature_process` to work on some videos from December and test our buffer detection methods

sneakers-the-rat:
> Right now the minimum pixel is always 16

interesting... why is that? I would think that it would be 0 from [here](https://github.com/Aharoni-Lab/mio/blob/56397e9f5167b4836589d073ab89bd8df61478e8/mio/stream_daq.py#L212)

> @jonny how would you structure adding this black method differently?

In general, this is a more flexible pattern:

```python
class ThingDoer:

    @abstractmethod
    def do(self, data) -> result:
        ...

class Version1(ThingDoer):
    def do(self, data) -> result:
        # do it version 1's way
        
class Version2(ThingDoer):
    def do(self, data) -> result:
        # do it version 2's way
```

than this

```python
class ThingDoer:
    def do_version_1(self, data) -> result: ...
    
    def do_version_2(self, data) -> result: ...
```

for a few reasons, primarily that we then have a consistent calling pattern - "all Thing Doers behave the same way" and they have all their functionality decoupled: e.g. rather than checking if there are no values or accidentally messing up the processing from one method by changing another as in the 'process first frame' bug from last week, we avoid those problems in the first place as well as avoiding needing a dispatching method. it's also way clearer to end users than stacking them up into a single class - even if the config parameters are independent and we know that, to an end user looking at that they'll be like "which of these are relevant to what again?" as opposed to being like "ah yes the BlackNoiseDetection class takes these specific config values..." and this will also become v relevant when these get integrated into the pipelining system

t-sasatani:
> Right now the minimum pixel is always 16

Very interesting. It might be a new type of error or some glitch in casting, etc. I can't work on this today but I'll take a look later too.

> for what would we want to use this? just curious.

We had some discussions on Slack about how this could potentially help eliminate the stripes. Though it's unlikely that this will help, it still could be worth a try if this area exists.

t-sasatani:
I made some updates so multiple detection methods can be mixed and did some general refactoring around the process module. Reviewing this and the `feat-preprocess` branch separately doesn't make sense, as these are tightly coupled. So, I'll merge this as soon as the pixel baseline bug (if it is a bug) is resolved.

t-sasatani:
I looked into the video output from the `stream` module, and there are areas of `16` at that point. S, it's probably not a bug in this module. I just posted issue #101 and will get this merged into the base denoising branch.

PR #105, opened 2025-02-16
Add `update_freq_mask` method

t-sasatani:
This PR adds a method for updating the frequency mask during processing. If we want to do adaptive frequency masking in the `mio process denoise` module, adding `freq_mask_processor.update_freq_mask(params)` right before the following code (which processes each frame in the `for` loop) should be most straightforward.
https://github.com/Aharoni-Lab/mio/blob/993617163700b87e851cbd2f6c58c46ec55edbc7/mio/process/video.py#L553

----
📚 Documentation preview 📚: https://miniscope-io--105.org.readthedocs.build/en/105/

### DIFF
--- a/mio/process/frame_helper.py
+++ b/mio/process/frame_helper.py
@@ -355,7 +355,7 @@ class FrequencyMaskHelper(BaseSingleFrameHelper):
         self._height = height
         self._width = width
         self._freq_mask_config = freq_mask_config
-        self._freq_mask = self._gen_freq_mask()
+        self._freq_mask = None
 
     @property
     def freq_mask(self) -> np.ndarray:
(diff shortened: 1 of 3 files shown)


MarcelMB:
As I see it it basically just preparing the code for an adaptive filtering right? There is no mechanism for the mask to change based on feedback from the processed frames (like a noise estimation). It would still take the user pre-defined values for all frames.

mio/process/frame_helper.py
```diff
@@ -365,8 +365,32 @@ def freq_mask(self) -> np.ndarray:
...
+        """
+        Update the frequency mask.
+
+        Parameters:
+            vertical_BEF_cutoff (int): Cutoff for the vertical band elimination filter in px.
+            horizontal_BEF_cutoff (int): Cutoff for the horizontal band elimination filter in px.
+            spatial_LPF_cutoff_radius (int): Radius for the spatial low pass filter cutoff in px.
+        """
+
+        self._freq_mask = self._gen_freq_mask(
```
MarcelMB, line 388: does it make sense to add logging here:

logging.info(f"Updating frequency mask with vertical_BEF_cutoff={vertical_BEF_cutoff}, "
                     f"horizontal_BEF_cutoff={horizontal_BEF_cutoff}, "
                     f"spatial_LPF_cutoff_radius={spatial_LPF_cutoff_radius}")
                     
                   to see successfully updated values?

MarcelMB (approved):
other than my question and one comment I'm fine with it!

PR #107, opened 2025-02-21
Butter filter for mean intensity values

MarcelMB:
Correction for fluctuations of the brightness of the entire field of view

I think its pretty straightforward and can be easily merged. The filter is not absolutely necessary but would make the calcium traces hopefully a bit noisy (need to test this better)

- calculate the mean pixel intensity for every frame

- apply a butter low pass filter to mean intensity over time

- creates a figure using start_frame to end_frame config settings, showing raw data and butter filtered data to visually assess if the filter resembles the data

after pressing 'Esc' script continues running

this filter is applied to the already spatial mask-filtered data (if enabled)
similar method used in Daniels notebook from v4 Miniscope
ipynb notebook
he used it to: 'correct the other source of V4 Miniscope noise. This is a fast, ~+3Hz fluctuation of the brightness'

I also saw some higher frequency noise on the calcium traces of the wireless miniscope

invivo-20241203-1-003_trim9min_butter_filter_intensity_plot

----
📚 Documentation preview 📚: https://miniscope-io--107.org.readthedocs.build/en/107/

### DIFF
--- a/mio/data/config/process/denoise_example.yml
+++ b/mio/data/config/process/denoise_example.yml
@@ -22,8 +22,8 @@ noise_patch:
 frequency_masking:
   enable: true
   cast_float32: true
-  spatial_LPF_cutoff_radius: 15
-  vertical_BEF_cutoff: 2
+  spatial_LPF_cutoff_radius: 3
+  vertical_BEF_cutoff: 3
   horizontal_BEF_cutoff: 0
   output_mask: true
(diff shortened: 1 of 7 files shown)


MarcelMB:
closed an earlier pull request of this since the code was so far behind the feature_process changes and main pipeline organization changes that it was easier to just re-implement them into the current framework

PR #108, opened 2025-02-22
Save/load config settings

MarcelMB:
I assume this PR is going to be more controversial

- first I based it off butter_filter (my other PR I just opened), should have maybe started it from feature_process and not nest it
- second it is failing some tests 
RROR tests/test_io.py
ERROR tests/test_models/test_model_process.py
ERROR tests/test_stream_daq.py

and here I probably need help. Because I don't know how to integrate it exactly or whats best practice. 

1. The idea is since we now have a bunch of preprocessing steps we would need to keep track what values we used for a specific file to make the preprocessing repeatable. We have this values in denoise_example.yml. So I save the most important values in a json file (not the ones that are just for visualization)
2. we than want to have a CLI command to use this saved json file to update the parameters in denoise_example.yaml

`mio process denoise -i video.avi -c denoise_example -u 'relativepath to' processing_log.json`
the -u stand for update and uses the saved.json file

one can still run just the normal command without updating the config yaml and just using whatever its in there
`mio process denoise -i video.avi -c denoise_example'

this all works well on my end by just playing around with changing thing in the json and using -u to update it on a next run.
But I am falling test since changing the CLI probably doesn't make all the tests happy. But let me know

### DIFF
--- a/mio/cli/process.py
+++ b/mio/cli/process.py
@@ -2,10 +2,14 @@
 Command line interface for offline video pre-processing.
 """
 
+from pathlib import Path
+from typing import Optional
+
 import click
 
 from mio.models.process import DenoiseConfig
 from mio.process.video import denoise_run
+from mio.utils.config import update_config_from_json
 
 
 @click.group()
@@ -31,12 +35,29 @@ def process() -> None:
     type=str,
     help="Path to the YAML processing configuration file.",
 )
+@click.option(
+    "-u",
+    "--update-from-json",
+    type=click.Path(exists=True, path_type=Path),
+    help="Update config from a processing_log.json file",
+)
+@click.option(
+    "--save-yaml",
+    is_flag=True,
+    help="Save updated config back to denoise_example.yml",
+)
 def denoise(
     input: str,
     denoise_config: str,
+    update_from_json: Optional[Path] = None,
+    save_yaml: bool = False,
 ) -> None:
     """
     Denoise a video file.
     """
-    denoise_config_parsed = DenoiseConfig.from_any(denoise_config)
-    denoise_run(input, denoise_config_parsed)
+    config = DenoiseConfig.from_any(denoise_config)
+
+    if update_from_json:
+        config = update_config_from_json(config, update_from_json)
+
+    denoise_run(input, config)
--- a/mio/process/video.py
+++ b/mio/process/video.py
@@ -2,6 +2,8 @@
 This module contains functions for pre-processing video data.
 """
 
+import json
+from datetime import datetime
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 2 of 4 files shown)


t-sasatani:
I'm not sure if it's nice to have two formats that basically contain the same thing. Might be more straightforward to have a `yaml_export` method in the config model and save the YAML file with the export video (with a hash or identifier if necessary)?

MarcelMB:
`________________________________________________________ ERROR collecting tests/test_io.py ________________________________________________________
ImportError while importing test module '/Users/mbrosch/Documents/GitKraken_mac/miniscope-io/tests/test_io.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/homebrew/Cellar/python@3.11/3.11.11/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_io.py:15: in <module>
    from mio.utils import hash_file, hash_video
E   ImportError: cannot import name 'hash_file' from 'mio.utils' (/Users/mbrosch/Documents/GitKraken_mac/miniscope-io/mio/utils/__init__.py)
____________________________________________ ERROR collecting tests/test_models/test_model_process.py _____________________________________________
ImportError while importing test module '/Users/mbrosch/Documents/GitKraken_mac/miniscope-io/tests/test_models/test_model_process.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/homebrew/Cellar/python@3.11/3.11.11/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_models/test_model_process.py:6: in <module>
    from mio.utils import hash_video, hash_file
E   ImportError: cannot import name 'hash_video' from 'mio.utils' (/Users/mbrosch/Documents/GitKraken_mac/miniscope-io/mio/utils/__init__.py)
____________________________________________________ ERROR collecting tests/test_stream_daq.py ____________________________________________________
ImportError while importing test module '/Users/mbrosch/Documents/GitKraken_mac/miniscope-io/tests/test_stream_daq.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/homebrew/Cellar/python@3.11/3.11.11/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_stream_daq.py:15: in <module>
    from mio.utils import hash_video, hash_file
E   ImportError: cannot import name 'hash_video' from 'mio.utils' (/Users/mbrosch/Documents/GitKraken_mac/miniscope-io/mio/utils/__init__.py)`

The tests seem to fail for some hash functions. I am not sure how this could be related to the json file saving and loading functionality?!

MarcelMB:
> save the YAML file with the export video (with a hash or identifier if necessary if uniqueness/history is an issue)
so you mean instead of saving a json save a yaml?

i mean the file format is of course not that important. Just saving a file that can always live with the video files in a folder. to basically know what the config was or re-run it if necessary on the same file or other files by another person

t-sasatani:
Yes, I agree that it's nice to save config files, and the format doesn't matter. But my question is, why don't we just copy the config file into the export folder instead of manually dumping everything?

t-sasatani:
I guess it'll be ideal if we can throw this stuff or some metadata into the video container itself, but it seems tricky with `.avi`.

MarcelMB:
> make a model export that can be used directly as a config file
sure that makes sense. Would this file still live in the user output folder or when loded replace the actual denoise config yaml in the config folder?

for now I never change the actual config file. it stays how it is downloaded. But only temporarily updates from the json file. if -u is used otherwise does what ever is in there

t-sasatani:
I think the primary need for this is to record the configs with the output files. So I think the key dump @sneakers-the-rat posted is pretty much enough to cover short-term needs.

For the design of configs in general, the denoise and device configs probably have different needs. I'm not exactly sure what's a good interface that fits both:
- `denoise` configs are pretty dynamic. We're now playing around with filter parameters (cutoff, etc.), so numerous minor config adjustments exist. I'm unsure if we should have an ID for each, but I want to preserve raw processing parameters with the output. The denoising may require users to play around with parameters, even in the final form of mio. It might be better to interact with notebooks if that's something people are comfortable with, or a GUI.
- `device` configs are matched one-to-one with the device (or at least the firmware version), and will be much more static. So, I guess keeping the ID is usually sufficient.

sneakers-the-rat:
i think recording the configs with the output is a really good idea - i would just like to spec that out a bit more explicitly :) that is the primary place that downstream analysis and data format tools will interact with this package, so it's worth having that be explicit and declarative. e.g. see the nwb miniscope extension - it's very dependent on the structure of the config model and output directories: https://github.com/catalystneuro/ndx-miniscope/tree/main/src/pynwb/ndx_miniscope

> the denoise and device configs probably have different needs. I'm not exactly sure what's a good interface that fits both:

this is super true, and one of the reasons i decided it was worth the complexity to have an `extends` and `include`-style system in the configs. it seemed pretty clear that we have different kinds of configs, but all of them need to be editable, so it should be possible for someone to e.g. take one and just tweak it slightly without needing to modify or copy the slow-moving device-specific config. 

hard agree we need to have a GUI for this, and hopefully all the data modeling makes the GUI part a lot less work. a decent amount of what i've been thinking about re: making configs not care about paths is in prep for making GUIs on top of this package. 

just want to reiterate i totally understand and agree with the need for this PR, but think that we probably want to do it in a different way, and since i'm the one that's saying that i'm willing to put up the work to make that happen. i'll have increasing time to work on mio stuff in the coming weeks.

Issue #111, opened 2025-03-13
OpenCV makes minimum pixel values `16` for no reason


sneakers-the-rat:
Related to: https://github.com/Aharoni-Lab/mio/issues/32

@MarcelMB noticed this odd problem where all the videos have a minimum pixel value of 16 in the video output

![Image](https://github.com/user-attachments/assets/3cbf63a3-f379-4c3c-9277-73655e7eb5e7)

We went ahead and made a stripped down version of the processing pipeline to try and investigate where this happens: 

- https://github.com/MarcelMB/bit_file_analysis
- https://github.com/sneakers-the-rat/bit_file_analysis

and it turns out it happens when writing with opencv - the frame values are correct at every point up until then.

![Image](https://github.com/user-attachments/assets/110293c6-80af-49b7-bf30-2ad8681cebfc)

but when we used ffmpeg to write the videos no such problem was present:

```python
from skvideo.io import FFMpegWriter

def write_video_ffmpeg(frames):
    writer = FFmpegWriter("test_ffmpeg.avi", outputdict={
        "-vcodec": "rawvideo",
        "-f": "avi",
        "-filter:v": "format=gray"
    })
    for frame in frames:
        writer.writeFrame(frame)
    writer.close()
```

![Image](https://github.com/user-attachments/assets/b17df1d6-5b9a-4ed9-a3e7-278f54cc19fc)

conclusion: we need to swap out opencv with a wrapper around ffmpeg.

t-sasatani:
This is awesome. Answers many questions:
* #32
* #101 
* #104

t-sasatani:
This is great! Solves many of our confusions.

Would you mind quickly adding this before making the large move to pipelines, as I believe it's pretty critical for ongoing stuff? (or I'm totally happy to quickly add this tomorrow).
I prefer adding it to the #83 PR branch because that also decouples the `VideoWriter` class from `stream_daq`, but I guess that conflict will be minor, so wherever is good.

t-sasatani:
Fixed in #112

PR #112, opened 2025-03-18
ffmpeg-based `VideoWriter`

t-sasatani:
This PR primarily integrates the bugfixes discussed in:
- #111 

# Updates
1. **FFmpeg-based `VideoWriter` class**: Open-cv-based video writers cause problems, so this PR switches to FFMpeg-based stuff. I haven't changed the real-time video displays and `.png` export because I assume it's very unlikely these are accidentally compressed, like the video export.
2. **Byte/bit operation moved to preprocessing**: This is still probably not in the ideal place, but I moved it to the earliest point I could without coordinating with the firmware part.
3. Removed USART support (for now).

This PR comes from the `feat-preprocess` branch because the `VideoWriter` class was isolated there to use it in the preprocessing feature.

----
📚 Documentation preview 📚: https://miniscope-io--112.org.readthedocs.build/en/112/

### DIFF
--- a/mio/io.py
+++ b/mio/io.py
@@ -10,6 +10,7 @@ from typing import Any, BinaryIO, Iterator, List, Literal, Optional, Tuple, Unio
 
 import cv2
 import numpy as np
+from skvideo.io import FFmpegWriter
 from tqdm import tqdm
 
 from mio.exceptions import EndOfRecordingException, ReadHeaderException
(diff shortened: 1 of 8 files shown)


MarcelMB:
Screenshot 2025-03-18 at 10 49 12 AM
I can confirm with a miniscope recording that it works well.

MarcelMB (approved):
works well and changes make sense!

mio/stream_daq.py
```diff
@@ -143,38 +144,6 @@ def nbuffer_per_fm(self) -> int:
             self._nbuffer_per_fm = len(self.buffer_npix)
         return self._nbuffer_per_fm
 
-    def _parse_header(self, buffer: bytes) -> Tuple[StreamBufferHeader, np.ndarray]:
```
sneakers-the-rat, line 146: i don't really have a strong opinion, i'm just not sure why this method had to die? fine with me tho

mio/stream_daq.py
```diff
@@ -359,8 +282,19 @@ def _fpga_recv(
                             buf_stop + len(self.preamble),
                         )
                     try:
+                        header, serial_buffer = BufferFormatter.bytebuffer_to_ndarrays(
```
sneakers-the-rat, line 285: i think we *might* take a tiny perf hit here by moving the few array operations that are done in this method into the read loop, but i also ultimately think the multiprocessing apart from the read loop is overkill that makes working with this class harder than it needs to be, so i plan on splitting it up anyway and this is also fine to me if it doesn't break stuff for y'all

mio/io.py
```diff
@@ -21,50 +22,44 @@
...
-        fourcc: str = "Y800",
-        **kwargs: dict,
-    ) -> cv2.VideoWriter:
+        output_dict: dict = None,
+    ):
         """
-        Create a parameterized video writer
+        Initialize the VideoWriter object.
+        """
+        if output_dict is None:
```
sneakers-the-rat, line 37: good heavens, parameters!!!!! yes!!!

sneakers-the-rat (approved):
oh my god yes

mio/stream_daq.py
t-sasatani: Also related to the perf comment in the read loop, I wanted to isolate the numpy's vectorized operations from the pydantic because I had a bit of resistance in putting any pydantic in the read loop (not sure if it matters, but I recently had some trouble with the speed of pydantic stuff so just to be safe).

Pulling numpy operations out resulted in two lines, so this method died, but I guess this small method makes it easier to read. I'll move this back to something that grabs the header `ndarray` and returns the same before merging later.

mio/stream_daq.py
t-sasatani: Yes, I haven't extensively tested, but hopefully, it's not critical as it's only some partially vectorized numpy operations.

mio/stream_daq.py
sneakers-the-rat: fine with it either way :) yes definitely agree should only ever use `model_construct` with pydantic models in perf-sensitive areas, validation is a luxury reserved for slower data

t-sasatani: ok the tiny _parse_header method is back

t-sasatani:
Thanks for the reviews! Will merge.

PR #113, opened 2025-06-27
Processing gs frame data

sneakers-the-rat:
discussed in slack.

process the weird 10/12-bit data to 1d pixel arrays, and then reshape into a frame

----
📚 Documentation preview 📚: https://miniscope-io--113.org.readthedocs.build/en/113/


hsemwal:
Hey @sneakers-the-rat 
[test-165.csv](https://github.com/user-attachments/files/20988784/test-165.csv)

OK I think something is going on with the orientation of the byte processing. I was looking into the header values, and they are really large numbers (even when I restart the scope) that are generally static (not changing). I provided a copy of the .csv here

Could the config file not be properly getting read, specifically in MSUS.yml:
# Flags to flip bit/byte order when recovering headers and data. See model document for details.
reverse_header_bits: False # False
reverse_header_bytes: False
reverse_payload_bits: False # True
reverse_payload_bytes: False #  True

mio/devices/gs/header.py
```diff
@@ -55,10 +56,11 @@ def from_buffer(
     ) -> tuple[Self, np.ndarray]:
         """Split buffer into a :class:`.GSBufferHeader` and a 1D, 16-bit pixel array."""
         header_start = len(config.preamble) * config.dummy_words
+        print(header_start)
         header_end = header_start + (header_fmt.header_length * 4)
         header_array = np.frombuffer(buffer[header_start:header_end], dtype=np.uint32)
         header = cls.from_format(header_array, header_fmt, construct=True)
-        payload = buffer_to_array(buffer[header_end:])
+        payload = buffer_to_array(buffer[header_end+1:]) # added +1 for header end+1
```
sneakers-the-rat, line 72: python is zero-indexed, this is incorrect. try testing your assumptions in reduced examples (or writing tests!!!!!!!)

```python
>>> mylist = (["preamble"] * 3) + (["header"] * 5) + (["not header"] * 5)
>>> mylist
['preamble',
 'preamble',
 'preamble',
 'header',
 'header',
 'header',
 'header',
 'header',
 'not header',
 'not header',
 'not header',
 'not header',
 'not header']
>>> header_len = 5
>>> header_start = 3
>>> header_end = header_start + header_len
>>> mylist[header_start:header_end]
['header', 'header', 'header', 'header', 'header']
>>> mylist[header_end:]
['not header', 'not header', 'not header', 'not header', 'not header']
```

sneakers-the-rat:
also, again, write tests that isolate the code you are trying to fix. it *could* be the case that you need to swap bits or bytes, but because of the error in `header_end + 1`, you would not be able to tell! 

## remember: you have to diagnose a bug before you try and fix it

PR #115, opened 2025-07-01
Msus test data

hsemwal:
Here, I will implement a test using real data from a different scope

Proof of Not-Working

I don't have this implemented so it doesn't really work.
plus the lint tests come back with failures all the time
Steps to complete

capture a small (<=5MB) sample of raw binary data coming out of the new scope, add it to the tests/data directory
make a test fixture that creates a version of your StreamDaq with default configuration that reads from that .bin file - [see the “default_streamdaq” fixture for an example](https://github.com/Aharoni-Lab/mio/blob/975151f102c2baaec25c1e9c436226b96757ad12/tests/test_stream_daq.py#L20) and [read the pytest docs re: fixtures](https://docs.pytest.org/en/latest/how-to/fixtures.html)
Current Status

I recorded .bin data
now going to try* and implement the code using some of the pre-existing code

----
📚 Documentation preview 📚: https://miniscope-io--115.org.readthedocs.build/en/115/


sneakers-the-rat:
Btw, so next step after #116 was merged is to then update these branches so you can use it. The MSUS branches are a little bit of chaos right now, like you are pulling into a branch of a branch of a branch here. Tidy up as you go so you don't grow a maze.

 so like I said in the prior PR:

- decide on a branch to use as the "main" branch for MSUS stuff - I recommend `msus`
- merge all the stuff you want into that branch, delete the other branches
- merge `main` into that branch.

Then as you work...
- make new branch.
- do a small amount of work in it
- merge that into the `msus` (or whatever you want to call it) branch

Until we've gotten it basically working. Only reason we aren't merging into `main` is educational reasons - giving you side branches to practice this flow.

tests/test_gs/test_stream_new_scope.py
```diff
@@ -0,0 +1,39 @@
...
+    set_okdev_input(data_file)
+
+    output_file = tmp_path / "output.bin"
+
+    daq_inst = StreamDaq(device_config=daqConfig)
+    daq_inst.capture(source="fpga", binary=output_file, show_video=False)
+
+    assert output_file.exists()
+
+    assert hash_file(data_file) == hash_file(output_file)
```
sneakers-the-rat, line 39: Haha they fuckin better match

pyproject.toml
```diff
@@ -120,7 +120,7 @@ filterwarnings = [
 testpaths = [
     "tests"
 ]
-timeout = 60
+# timeout = 60
```
sneakers-the-rat, line 123:
```suggestion
timeout = 60
```

I am uncommenting this because this PR keeps taking up all the ci runner slots because of breakpoints left in committed code and other non-termination problems.

sneakers-the-rat:
looks like this is stale, let's chat about the path towards merging the MSUS stuff :)

PR #116, opened 2025-07-02
Separate buffer-splitting in `_fpga_recv` into a generator function

sneakers-the-rat:
Doing a live programming demo with @hsemwal , and for ~secret miniscope type~ we want to be able to test the pixel-wise and frame-wise processing of the data outside of the streamdaq class as we get it on its feet. since it's pretty hard to do that as is, this is a function-neutral change to just split out the iteration (as it would be done in the guts of the streamdaq) into a separate function.

i'm not sure how much we've actually said this on this repo, but all the pipelining stuff is getting split off into a separate package because it felt sort of absurd to do it all in `mio`, so until that's done, will be making smaller more targeted refactors.

should be *relatively* self explanatory, the one thing that might not be is making the stream device into an iterator with `__next__` and `__iter__` - we make the simplifying assumption here that we dont' want to change the read size mid-iteration in normal circumstances, but preserve the ability to do that by keeping the `length` param in the `okDev.read_data` method as optional. that lets us make the `iter_buffers` function generic: any iterator that yields bytes can be passed, and all it does it yields bytes after combining them and splitting by the `prefix`. if we want that to be more specialized (e.g. fuzzy prefix matching #42 ) we can make that into a class, but for now the generator is a way to have a tiny bit of state in a predictable way.

should be a function-neutral change, just making it to help ~secret miniscope type~ development in downstream branch while doing a programming lesson in small, self-contained PRs 

----
📚 Documentation preview 📚: https://miniscope-io--116.org.readthedocs.build/en/116/

### DIFF
--- a/tests/test_devices/test_okdev.py
+++ b/tests/test_devices/test_okdev.py
@@ -1,0 +1,42 @@
+import pytest
+from pathlib import Path
+from mio.devices.mocks import okDevMock
+
+
+@pytest.fixture
+def _tmp_buffer(tmp_path: Path, set_okdev_input) -> bytes:
+    buffer = b"12345"
+    buffer_rep = b"12345" * 3
+    data_path = tmp_path / "data.bin"
+    with open(data_path, "wb") as f:
+        f.write(buffer_rep)
+    set_okdev_input(data_path)
+    return buffer
+
+
+def test_okdev_iter(_tmp_buffer: bytes) -> None:
+    """
+    sanity check that the __iter__ method for iteration on the okdev
+    does iter things
+    """
+    dev = okDevMock(read_length=5)
+    got_buffers = 0
+    for buf in dev:
+        assert buf == _tmp_buffer
+        got_buffers += 1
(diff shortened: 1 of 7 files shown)


tests/test_devices/test_okdev.py
```diff
@@ -0,0 +1,42 @@
...
+def test_okdev_next(_tmp_buffer: bytes) -> None:
+    """
+    the okDevMock.__next__ method (and by proxy okDev) does next things
+    """
+    dev = okDevMock(read_length=5)
+    buffers = []
+    buffers.append(next(dev))
+    buffers.append(next(dev))
+    buffers.append(next(dev))
+    with pytest.raises(StopIteration):
```
sneakers-the-rat, line 39: i don't think i explained this in the video call - @hsemwal what this means is that the test fails if the error is *not* raised here

sneakers-the-rat:
merging this without waiting for review since it doesn't change the way anything works and is just to provide ability to test for now!

PR #120, opened 2025-07-22
Unify Frame/Video models for different I/Os

t-sasatani:
This shouldn't change the actual functions.
- (Minimally) fused the Frame/Video models for different I/Os.
- Validate frame/video models using `numpydantic`
- Made an IO module with separate files for file and SDcard I/O (primarily for avoiding circular imports).

----
📚 Documentation preview 📚: https://miniscope-io--120.org.readthedocs.build/en/120/

### DIFF
--- a/pylock.toml
+++ b/pylock.toml
@@ -590,6 +590,23 @@ dependencies = [
 ]
 
 [[packages]]
+name = "numpydantic"
+version = "1.8.0"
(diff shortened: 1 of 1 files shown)


t-sasatani:
It's behind main quite a bit so I'll come back to this later

sneakers-the-rat (approved):
hell ya

PR #121, opened 2025-07-22
Online frequency filter (stripe removal)

t-sasatani:
Online (real-time) bandpass filtering integrated with the `stream` command.
```
mio stream capture -c wireless-200px -f remove_stripe_example
```
This only filters the video preview, leaving the file output as is.

----
📚 Documentation preview 📚: https://miniscope-io--121.org.readthedocs.build/en/121/

### DIFF
--- a/mio/data/config/process/denoise_example.yml
+++ b/mio/data/config/process/denoise_example.yml
@@ -20,6 +20,9 @@ noise_patch:
   output_diff: true
   output_noisy_frames: true
 frequency_masking:
+  id: frequency_masking_example
+  mio_model: mio.models.process.FreqencyMaskingConfig
+  mio_version: 0.6.1
   enable: true
   cast_float32: true
   spatial_LPF_cutoff_radius: 15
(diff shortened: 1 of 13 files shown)


mio/data/config/process/denoise_example.yml
```diff
@@ -20,6 +20,9 @@ noise_patch:
   output_diff: true
   output_noisy_frames: true
 frequency_masking:
+  id: frequency_masking_example
+  mio_model: mio.models.process.FreqencyMaskingConfig
+  mio_version: 0.6.1
   enable: true
```
t-sasatani, line 26: It's somewhat awkward having to do this to make the FrequencyMaskingConfig an actual configuration that can also work independently, but it's also convenient to have all configurations in a single file.

Should I make this a separate file, or could (should) we possibly change the ConfigYAMLMixin to make the ID only necessary for non-nested configs?

t-sasatani:
The coverage is decreasing because the CLI file has not been tested. Should we test this at all? (It's zero coverage now)

PR #125, opened 2025-07-30
Fix csvwriter header alignment, fix writing duplicated header rows

sneakers-the-rat:
@hsemwal reported an issue where the header metadata looked correct when it was being printed to stdout, but the csv being written was incorrect s.t. each buffer was being written multiple times....

<details>
<summary>expand/collapse example</summary>

|linked_list|frame_num                    |buffer_count|frame_buffer_count                           |write_buffer_count|dropped_buffer_count|timestamp|pixel_count|write_timestamp|battery_voltage_raw|input_voltage_raw|unix_time         |
|-----------|-----------------------------|------------|---------------------------------------------|------------------|--------------------|---------|-----------|---------------|-------------------|-----------------|------------------|
|11         |12821                        |102571      |3                                            |102570            |0                   |624079   |5072       |0              |0                  |0                |1753833136.014915 |
|30         |12823                        |102590      |6                                            |102587            |0                   |624183   |5072       |0              |0                  |0                |1753833136.0149791|
|11         |12821                        |102571      |3                                            |102570            |0                   |624079   |5072       |0              |0                  |0                |1753833136.02576  |
|30         |12823                        |102590      |6                                            |102587            |0                   |624183   |5072       |0              |0                  |0                |1753833136.0258172|
|31         |12823                        |102591      |7                                            |102588            |0                   |624185   |4496       |0              |0                  |0                |1753833136.025866 |
|11         |12821                        |102571      |3                                            |102570            |0                   |624079   |5072       |0              |0                  |0                |1753833136.21313  |
|30         |12823                        |102590      |6                                            |102587            |0                   |624183   |5072       |0              |0                  |0                |1753833136.213239 |
|31         |12823                        |102591      |7                                            |102588            |0                   |624185   |4496       |0              |0                  |0                |1753833136.213291 |
|0          |12824                        |102592      |0                                            |102592            |0                   |624218   |5072       |0              |0                  |0                |1753833136.213341 |
|1          |12824                        |102593      |1                                            |102593            |0                   |624220   |5072       |0              |0                  |0                |1753833136.213973 |
|2          |12824                        |102594      |2                                            |102593            |0                   |624222   |5072       |0              |0                  |0                |1753833136.2140312|
|3          |12824                        |102595      |3                                            |102594            |0                   |624225   |5072       |0              |0                  |0                |1753833136.214083 |
|4          |12824                        |102596      |4                                            |102594            |0                   |624227   |5072       |0              |0                  |0                |1753833136.214133 |
|1          |12824                        |102593      |1                                            |102593            |0                   |624220   |5072       |0              |0                  |0                |1753833136.217022 |
|2          |12824                        |102594      |2                                            |102593            |0                   |624222   |5072       |0              |0                  |0                |1753833136.217081 |
|3          |12824                        |102595      |3                                            |102594            |0                   |624225   |5072       |0              |0                  |0                |1753833136.217134 |
|4          |12824                        |102596      |4                                            |102594            |0                   |624227   |5072       |0              |0                  |0                |1753833136.217185 |
|5          |12824                        |102597      |5                                            |102595            |0                   |624230   |5072       |0              |0                  |0                |1753833136.217236 |
|1          |12824                        |102593      |1                                            |102593            |0                   |624220   |5072       |0              |0                  |0                |1753833136.219507 |
|2          |12824                        |102594      |2                                            |102593            |0                   |624222   |5072       |0              |0                  |0                |1753833136.2195652|
|3          |12824                        |102595      |3                                            |102594            |0                   |624225   |5072       |0              |0                  |0                |1753833136.2196171|
|4          |12824                        |102596      |4                                            |102594            |0                   |624227   |5072       |0              |0                  |0                |1753833136.219667 |
|5          |12824                        |102597      |5                                            |102595            |0                   |624230   |5072       |0              |0                  |0                |1753833136.219718 |
|6          |12824                        |102598      |6                                            |102595            |0                   |624232   |5072       |0              |0                  |0                |1753833136.219767 |
|1          |12824                        |102593      |1                                            |102593            |0                   |624220   |5072       |0              |0                  |0                |1753833136.2221272|
|2          |12824                        |102594      |2                                            |102593            |0                   |624222   |5072       |0              |0                  |0                |1753833136.222185 |
|3          |12824                        |102595      |3                                            |102594            |0                   |624225   |5072       |0              |0                  |0                |1753833136.2222688|
|4          |12824                        |102596      |4                                            |102594            |0                   |624227   |5072       |0              |0                  |0                |1753833136.2223349|
|5          |12824                        |102597      |5                                            |102595            |0                   |624230   |5072       |0              |0                  |0                |1753833136.222386 |
|6          |12824                        |102598      |6                                            |102595            |0                   |624232   |5072       |0              |0                  |0                |1753833136.2224362|
|7          |12824                        |102599      |7                                            |102596            |0                   |624234   |4496       |0              |0                  |0                |1753833136.222485 |
|1          |12824                        |102593      |1                                            |102593            |0                   |624220   |5072       |0              |0                  |0                |1753833136.304621 |
|2          |12824                        |102594      |2                                            |102593            |0                   |624222   |5072       |0              |0                  |0                |1753833136.304684 |
|3          |12824                        |102595      |3                                            |102594            |0                   |624225   |5072       |0              |0                  |0                |1753833136.3048398|
|4          |12824                        |102596      |4                                            |102594            |0                   |624227   |5072       |0              |0                  |0                |1753833136.304956 |
|5          |12824                        |102597      |5                                            |102595            |0                   |624230   |5072       |0              |0                  |0                |1753833136.305024 |
|6          |12824                        |102598      |6                                            |102595            |0                   |624232   |5072       |0              |0                  |0                |1753833136.305075 |
|7          |12824                        |102599      |7                                            |102596            |0                   |624234   |4496       |0              |0                  |0                |1753833136.305173 |
|8          |12825                        |102600      |0                                            |102600            |0                   |624266   |5072       |0              |0                  |0                |1753833136.3052218|

</details>

This was caused because we were passing the list of headers through when `buffer_npix` does not match the data being given by the device.

https://github.com/Aharoni-Lab/mio/blob/78870beec5edb11d67c1cec2fbf0d30aeab72572/mio/stream_daq.py#L355-L372

While trying to write the correctly failing test, i also noticed that the csv writer was just broken in general because there wasn't a good match between the header format and the writing of rows - the header row was coming from the `StreamBufferHeader` which is the class to contain *header data*, rather than the `StreamBufferHeaderFormat` which is the class that defines how the header is structured in the received buffers. 

as a result the headers were 
a) too short, so there were 12 columns of data but only 10 headers, and 
b) incorrect: the contents of each column were not actually the data they were supposed to be.

So i fixed that by making the `BufferedCSVWriter` have a `header` param (i don't think we ever want to write csvs that don't have a header row...). that defines the order and the columns, so then we pass individual rows as dictionaries - we only worry about the order of the columns at creation time, and that should determine the output structure of the whole csv. it should not be possible to accidentally break the table by passing an additional value to some row or not having some value.

---

the original bug is *not* a noisy signal bug, but an incorrect software or config problem - the `buffer_npix` is entirely determined by config, so if it doesn't match the number of buffers per frame being returned by the device, we should we made loudly aware of this.

First commit should be a correctly failing test confirming that i replicated the bug in the regression test, and then second should fix the test by simply not passing the header list multiple times into the queue.

----
📚 Documentation preview 📚: https://miniscope-io--125.org.readthedocs.build/en/125/

### DIFF
--- a/mio/io.py
+++ b/mio/io.py
@@ -188,7 +188,6 @@ class BufferedCSVWriter:
             keys that are not in `header` are ignored, and missing keys are `None`
         """
         row = [data.get(key) for key in self.header]
-        row = [int(value) if isinstance(value, np.generic) else value for value in row]
         self.buffer.append(row)
         if len(self.buffer) >= self.buffer_size:
             self.flush_buffer()
--- a/mio/io.py
+++ b/mio/io.py
@@ -6,7 +6,7 @@ import atexit
 import contextlib
 import csv
 from pathlib import Path
-from typing import Any, BinaryIO, Iterator, List, Literal, Optional, Tuple, Union, overload
+from typing import BinaryIO, Iterator, Literal, Optional, Tuple, Union, overload
 
 import cv2
 import numpy as np
@@ -147,6 +147,9 @@ class BufferedCSVWriter:
     ----------
     file_path : Union[str, Path]
         The file path for the CSV file.
+    headers : list[str]
+        Headers for csv - determine the order of columns and allowable values
+        passed to :meth:`.append`
     buffer_size : int, optional
         The number of rows to buffer before writing to the file (default is 100).
 
@@ -160,26 +163,33 @@ class BufferedCSVWriter:
         The buffer for storing rows before writing.
     """
 
-    def __init__(self, file_path: Union[str, Path], buffer_size: int = 100):
+    def __init__(self, file_path: Union[str, Path], header: list[str], buffer_size: int = 100):
         self.file_path: Path = Path(file_path)
+        self.header = header
         self.buffer_size = buffer_size
         self.buffer = []
         self.logger = init_logger("BufferedCSVWriter")
 
+        # write header in first row
+        self.buffer.append(self.header)
+
         # Ensure the buffer is flushed when the program exits
         atexit.register(self.flush_buffer)
 
-    def append(self, data: List[Any]) -> None:
+    def append(self, data: dict) -> None:
         """
         Append data (as a list) to the buffer.
 
         Parameters
         ----------
-        data : List[Any]
+        data : dict
             The data to be appended.
+            Rows are constructed and columns are ordered according to `header` -
+            keys that are not in `header` are ignored, and missing keys are `None`
         """
-        data = [int(value) if isinstance(value, np.generic) else value for value in data]
-        self.buffer.append(data)
+        row = [data.get(key) for key in self.header]
+        row = [int(value) if isinstance(value, np.generic) else value for value in row]
+        self.buffer.append(row)
         if len(self.buffer) >= self.buffer_size:
             self.flush_buffer()
 
--- a/tests/test_stream_daq.py
+++ b/tests/test_stream_daq.py
@@ -12,6 +12,7 @@ import time
 from contextlib import contextmanager
 from bitstring import BitArray, Bits
 from typing import Generator
+import warnings
 
 from mio import BASE_DIR
 from mio.stream_daq import StreamDevConfig, StreamDaq, iter_buffers
@@ -128,6 +129,35 @@ def test_csv_output(tmp_path, default_streamdaq, write_metadata, caplog):
     else:
         default_streamdaq.capture(source="fpga", metadata=None, show_video=False)
         assert not output_csv.exists()
+
+def test_processing_speed(tmp_path, default_streamdaq):
+    """
+    Processing speed test of the stream daq.
+    For being generous for runs in CI, the test will pass if the processing speed is faster than the test_fail_fps.
+    This will output a warning if it is slower than the warning_fps.
+    """
+    test_fail_fps = 10
+    warning_fps = 40
+    output_csv = tmp_path / "output.csv"
+
+    default_streamdaq.capture(source="fpga", metadata=output_csv, show_video=False)
+
+    df = pd.read_csv(output_csv)
+
+    unix_time_first = df.iloc[0]['unix_time']
+    unix_time_last = df.iloc[-1]['unix_time']
+    time_taken = unix_time_last - unix_time_first
+
+    frame_index_first = df.iloc[0]['frame_num']
+    frame_index_last = df.iloc[-1]['frame_num']
+    num_frames = frame_index_last - frame_index_first
+
+    processing_fps = num_frames / time_taken
+
+    if processing_fps < warning_fps:
+        warnings.warn(f"Processing speed is {processing_fps} FPS, which is slower than the required {warning_fps} FPS")
+
+    assert processing_fps > test_fail_fps
 
 def test_csv_no_duplicates(tmp_path, set_okdev_input):
     """
--- a/tests/test_stream_daq.py
+++ b/tests/test_stream_daq.py
@@ -2,6 +2,7 @@ import re
 from pathlib import Path
 
 import multiprocessing
+import numpy as np
 import os
 import pytest
 import pandas as pd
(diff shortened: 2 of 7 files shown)


mio/io.py
```diff
@@ -160,26 +163,33 @@ class BufferedCSVWriter:
...
-        data : List[Any]
+        data : dict
             The data to be appended.
+            Rows are constructed and columns are ordered according to `header` -
+            keys that are not in `header` are ignored, and missing keys are `None`
         """
-        data = [int(value) if isinstance(value, np.generic) else value for value in data]
-        self.buffer.append(data)
+        row = [data.get(key) for key in self.header]
+        row = [int(value) if isinstance(value, np.generic) else value for value in row]
```
t-sasatani, line 191: This conversion looks unnecessary, as the list always comes from a pydantic model now (with native Python number types, not `np` types), and anyway, casting `np.generic` to `int` seems extremely broad.

I think I initially added this conversion to handle `np.uintX`, but I guess this should be handled in the model.

```suggestion
```

t-sasatani (approved):
This looks great and is much more robust than before. I suggested one change (loosely related to this update), so please have a quick look.

mio/io.py
sneakers-the-rat: i *believe* we use `model_construct` for the headers since they are in a perf-sensitive loop, so in this case i think we do need the int coercion, but we should test this in any case. could be the case that at some point all the coercion outside the pydantic validation becomes more expensive than just using the validation directly, seems worth profiling.

t-sasatani: Ah, thanks, I overlooked the `construct=True` in `_parse_header` because this part isn't as perf-sensitive as the data receiving process (the header construction is in a process that grabs buffers from the queue populated by the hardware interfacing process).

I think it's fine to use Pydantic validation for this, as it's not in the first FPGA loop (maybe better to profile, though I'm not confident about doing it myself).

t-sasatani: This mostly felt confusing to me because this CSV writer also handles float values like UNIX time (we want sub-sec resolution).

sneakers-the-rat:
@t-sasatani if you test this new version and it seems to still be keeping up with the frames, that's good enough for me!

sneakers-the-rat:
oh nice! sorry i just meant run it with your device and see if it seems fine, but probably good to have a perf threshold test like that, thanks for writing, sorry if it took you out of your way :)

tests/test_stream_daq.py
```diff
@@ -129,6 +130,37 @@ def test_csv_output(tmp_path, default_streamdaq, write_metadata, caplog):
...
+    """
+    Processing speed test of the stream daq.
+    For being generous for runs in CI, the test will pass if the processing speed is faster than the test_fail_fps.
+    This will output a warning if it is slower than the warning_fps.
+    """
+    test_fail_fps = 10
+    warning_fps = 40
+    output_csv = tmp_path / "output.csv"
+
+    default_streamdaq.capture(source="fpga", metadata=output_csv, show_video=video_preview)
```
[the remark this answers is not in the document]
sneakers-the-rat: (that's where the segfaults are coming from, i believe)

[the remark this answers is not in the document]
t-sasatani: Ah, thanks, makes sense. I force pushed the update.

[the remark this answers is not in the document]
t-sasatani: Hmm, it seems something different. Sry in advance for the many test fail notifications in case it's coming.

t-sasatani:
Ok, the test failure was a mixed issue of the preview and the outdated apt for the ffmpeg install in Ubuntu.

I added a minimal performance test for `streamdaq`, which fails if the processing from binary is under 10 FPS (to be generous to CI runners) and shows a warning if it's under 40 FPS. These are sort of arbitrary, but should work as an alert when we add something significantly slow.

Also, it's quite interesting that the processing seems slightly faster with `construct=False` when I run on my local. Anyways, it seems ok to run header validations in Pydantic.

t-sasatani:
@sneakers-the-rat Thanks for finding the CI preview thing.  Pls feel free to merge if it looks alright. Or I can do it otherwise.

sneakers-the-rat:
> it's quite interesting that the processing seems slightly faster with construct=False when I run on my local.

what on earth. that is truly mysterious. well, in any case, this version is a win-win if it's both faster *and* we get validation. will merge now

sneakers-the-rat:
Added some changelog notes and updated the readme to make it clearer that this software isn't ready for external use. merging now and will cut a new release

PR #128, opened 2025-08-02
fix buffer_npix method in gs

sneakers-the-rat:
man what even am i looking at

### DIFF
--- a/tests/test_gs/test_header.py
+++ b/tests/test_gs/test_header.py
@@ -68,6 +68,7 @@ def test_buffer_npix(gs_raw_buffers):
     frames = list(frame_buffers.keys())
     del frame_buffers[frames[0]]
     del frame_buffers[frames[-1]]
+    breakpoint()
     for frame_num, pixels in frame_buffers.items():
         pixel_lengths = [len(p) for p in pixels]
         assert pixel_lengths == config.buffer_npix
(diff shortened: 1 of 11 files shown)


sneakers-the-rat (commented):
cool. so this should both merge the buffer_npix fixes as well as get the base branch partially back to a state it can be worked on. if we clean up as we go then we avoid piling a bunch of broken code on broken code and can actually make progress. if we just keep jamming our shit with a bunch of comments, tweaking things in a nonsystematic way, then that's how we get this project which should have taken a week, tops, taking >8 months.

tests/test_gs/test_header.py
```diff
@@ -60,16 +53,30 @@ def test_format_headers_raw(gs_raw_buffers):
...
+        frame_buffers[header.frame_num].append(pixels)
+
+    # discard first and last which may be incomplete in the sample
+    frames = list(frame_buffers.keys())
+    del frame_buffers[frames[0]]
+    del frame_buffers[frames[-1]]
+
+    for frame_num, pixels in frame_buffers.items():
+        pixel_lengths = [len(p) for p in pixels]
+        assert pixel_lengths == config.buffer_npix
```
sneakers-the-rat, line 73: this is the only thing that this PR is about really, the rest is just getting the code to run again

PR #133, opened 2025-09-11
Feature - stitching multiple video streams

t-sasatani:
This PR adds a stitching function for multiple data streams based on device metadata. Also makes frame index operations rigorous to support this.

The metadata CSV for testing stitching is also inflating the line count, but I'm sorry that this ended up quite long.

# Functions
1. Load multiple pairs of videos (.avi) and metadata (.csv) from parallel recording streams.
2. Align frames across recordings using frame_num from metadata.
3. For each aligned frame, score candidates to pick the best version:
    - Metadata scoring: prefer frames with more buffers and less black padding.
    - Image scoring (tiebreaker): use Sobel edge detection to pick the least sharp frame (high edge response likely indicates artifacts/noise).
4. Reconstruct the selected frames into a single stitched video and metadata file, with reconstructed_frame_index renumbered sequentially.
5. Optionally export debug outputs (composite comparison video + CSV with per-frame selection decisions, diff pixel counts, and edge scores).
6. Supports a full workflow pipeline (mio process workflow): stitch → crop (trim start/end frames) → denoise, with upfront validation and frame-count alignment checks at each step.

# Main updates
- `mio/process/stitch.py`: Main part that scores multiple recording videos from a bundle, gets the best ones, and exports while tracking and adjusting the metadata too.
- `mio/process/video.py`: Made the frame index tracking much more rigorous and better path resolving for getting a data bundle after running.
- `mio/cli/process.py`: Add `workflow` and `stitch` CLI commands alongside the existing `denoise`, add interactive prompts to proceed/quit when metadata validation fails.
- `mio/models/stitch.py`: Model for tracking frames from multiple video sources.
- `mio/utils.py`: Added some validation functions and formatters for video-metadata pairs.
- `tests/test_process/test_stitch.py`: Testing for stitching and cropping frames. Tests stuff like alignment of video and metadata, video hash, etc.

----
📚 Documentation preview 📚: https://miniscope-io--133.org.readthedocs.build/en/133/


sneakers-the-rat:
will be back at work tomorrow, lmk if this is ready to take a look at!

mio/cli/process.py
```diff
@@ -31,12 +74,432 @@ def process() -> None:
...
+        except ValueError as e:
+            raise click.ClickException(str(e)) from None
+
+        try:
+            validate_video_metadata_match(actual_cropped_video)
+        except VideoMetadataError as e:
+            raise click.ClickException(
+                f"Frame count alignment failed after cropping: {e}"
+            ) from None
+        click.echo(f"✅ [crop] Frame count alignment verified: {actual_cropped_video}")
```
sneakers-the-rat: same thing here, just call `trim()`

mio/models/stitch.py
```diff
@@ -0,0 +1,74 @@
...
+        if rfi_values.nunique() == 1:
+            reconstructed_frame_index = rfi_values.iloc[0]
+        else:
+            reconstructed_frame_index = rfi_values.mode()[0]
+            msg = (
+                f"Reconstructed frame index is not the same "
+                f"for all buffers in frame {frame_num}. "
+                f"Using the majority reconstructed_frame_index: {reconstructed_frame_index}"
+            )
+            tqdm.write(msg)
```
sneakers-the-rat: what's this doing?

mio/process/video.py
```diff
@@ -138,33 +143,41 @@ def __init__(
...
-                logger.info(f"Dropping frame {index} of original video due to noise.")
-                logger.debug(f"Adding noise patch for frame {index}.")
+                msg = f"Dropping frame {original_frame_index} of original video due to noise."
+                tqdm.write(msg)
+                logger.debug(msg)
+                logger.debug(f"Adding noise patch for frame {original_frame_index}.")
                 self.noise_patchs.append((noisy_area * np.iinfo(np.uint8).max).astype(np.uint8))
                 self.noisy_frames.append(input_frame)
-                self.dropped_frame_indices.append(index)
+                self.dropped_frame_indices.append(original_frame_index)
```
sneakers-the-rat: i am not really understanding the change here + addition of this param. it looks like the prior impl is pretty implicit so receiving this value rather than inferring it from internal state seems better, but i am not quite getting it. maybe the name is getting in the way of my understanding? could this param just be called `index`? what does the "original" part mean?

mio/process/video.py
```diff
@@ -529,39 +564,86 @@ def denoise_run(
     if config.interactive_display.display_freq_mask:
         freq_mask_processor.freq_mask_named_frame.display()
 
+    total_frames = int(reader.cap.get(cv2.CAP_PROP_FRAME_COUNT))
+
     try:
-        for index, frame in reader.read_frames():
+        frame_iter = tqdm(reader.read_frames(), total=total_frames, desc="Processing frames")
```
sneakers-the-rat: make pbars optional

mio/process/video.py
```diff
@@ -529,39 +564,86 @@ def denoise_run(
...
+            output_dir=intermediate_dir,
             video_frames=output_frames,
             minimum_projection_config=config.minimum_projection,
         )
         minimum_projection_processor.normalize_stack()
 
-        noise_patch_processor.batch_export_videos()
+        noise_patch_processor.export_output_video()
+
+        noise_patch_processor.output_dir = intermediate_dir
```
sneakers-the-rat: why is this assigned here rather than when the object is created?

mio/process/video.py
```diff
@@ -587,3 +669,282 @@ def denoise_run(
                 end_frame=config.interactive_display.end_frame,
             )
             video_plotter.show()
+
+    return output_video_path
+
+
+def _modify_csv_metadata(
+    input_video_path: str,
```
sneakers-the-rat: why is one a str and the other a path?

mio/process/video.py
```diff
@@ -587,3 +669,282 @@ def denoise_run(
...
+    try:
+        frame_timestamps.to_csv(output_csv_path, index=False)
+        logger.info(
+            "Successfully exported frame-timestamp CSV at %s (%d frames).",
+            output_csv_path,
+            len(frame_timestamps),
+        )
+    except Exception as e:
+        logger.error(f"Failed to write frame-timestamp CSV file {output_csv_path}: {e}")
+        raise
```
sneakers-the-rat: along with comments above, since writing to a file is just a one-liner, i generally like to separate operations that mutate arrays/dataframes from reading/writing them.

mio/process/video.py
```diff
@@ -587,3 +669,282 @@ def denoise_run(
...
+    input_path = Path(video_path)
+
+    if output_path is None:
+        output_path_obj = input_path.parent / f"{input_path.stem}_cropped{input_path.suffix}"
+    else:
+        output_path_obj = Path(output_path).expanduser()
+        if output_path_obj.is_dir() or not output_path_obj.suffix:
+            if not output_path_obj.exists():
+                output_path_obj.mkdir(parents=True, exist_ok=True)
+            output_path_obj = output_path_obj / f"{input_path.stem}_cropped{input_path.suffix}"
```
sneakers-the-rat: this happens in a bunch of places, i think pretty much everywhere except entrypoint functions we should just accept the paths we are going to work on as params and not do any modification otherwise. like there should be ideally only one place where the directory structure of a dataset is declared, rather than baked into every function

mio/process/video.py
```diff
@@ -587,3 +669,282 @@ def denoise_run(
...
+            if index < start_idx:
+                continue
+            if index > end_idx:
+                break
+
+            if len(frame.shape) != 2:
+                raise ValueError(f"Frame {index} has shape {frame.shape}, expected 2D grayscale.")
+
+            writer.write_frame(frame)
+            frames_written += 1
```
sneakers-the-rat: i think this should work like

```python
with VideoReader(...) as reader, VideoWriter(...) as writer:
    for frame in reader.iter_frames(start, stop):
        writer.write_frame(frame)
```

- handle releasing/closing with contextmanagers
- accept range as params, seek to start position. iterating through all frames and throwing them away at the start takes a lot of disk io and computation to decode the video.
- allow the writer and reader to handle throwing errors if shape is invalid/changes.

mio/process/video.py
```diff
@@ -587,3 +669,282 @@ def denoise_run(
...
+        f"({len(df_filtered)} rows)."
+    )
+
+    try:
+        df_filtered.to_csv(output_csv_path, index=False)
+        logger.info(f"Successfully modified CSV metadata at {output_csv_path}.")
+        return df_filtered
+    except Exception as e:
+        logger.error(f"Failed to write output CSV file {output_csv_path}: {e}")
+        raise
```
sneakers-the-rat: this seems like it should be 

```python
_modify_csv_metadata(..., 
    dropped_frame_indices=set(range(start_idx)) | set(range(end_idx, total_frames))
)
```

mio/exceptions.py
```diff
@@ -2,6 +2,13 @@
 Custom exceptions!
 """
 
+from __future__ import annotations  # needed for pd.DataFrame | None on <3.10
```
sneakers-the-rat:
```suggestion
from __future__ import annotations
```

no need to explain :)

mio/utils.py
```diff
@@ -63,3 +64,119 @@ def hash_video(
...
+        raise VideoMetadataError(
+            f"CSV file {csv_path_obj} does not have 'reconstructed_frame_index' column",
+            csv_df=df,
+        )
+
+    cap = cv2.VideoCapture(str(video_path_obj))
+    if not cap.isOpened():
+        raise VideoMetadataError(f"Could not open video file {video_path_obj}", csv_df=df)
+    frame_count = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+    cap.release()
```
sneakers-the-rat:
```suggestion
    cap = VideoReader(video_path_obj).frame_count
```

mio/utils.py
```diff
@@ -63,3 +64,119 @@ def hash_video(
...
+    video_path_obj = Path(video_path)
+    csv_path_obj = video_path_obj.with_suffix(".csv")
+
+    if not csv_path_obj.exists():
+        raise VideoMetadataError(f"CSV file not found at {csv_path_obj}")
+
+    try:
+        df = pd.read_csv(csv_path_obj)
+    except Exception as e:
+        raise VideoMetadataError(f"Failed to read CSV file {csv_path_obj}: {e}") from e
```
sneakers-the-rat:
```suggestion
    df = pd.read_csv(csv_path_obj)
```

these exceptions should happen just as a normal part of reading the csv and i dont' think we should handle them specially

mio/cli/process.py
```diff
@@ -31,12 +74,432 @@ def process() -> None:
...
+            validation_failures.append((video_path_obj, e))
+
+    if validation_failures:
+        click.echo("\n⚠️  Frame number validation found mismatches:")
+        click.echo("=" * 70)
+
+        for video_path_obj, error in validation_failures:
+            click.echo(f"\n📹 Video: {video_path_obj.name}")
+
+            details = extract_mismatch_details(video_path_obj, error.csv_df)
```
sneakers-the-rat: this is super strange to me - we basically are recalculating the errors after detecting that we have errors? why not just have the information about the exception in the exception itself?

mio/utils.py
```diff
@@ -63,3 +64,119 @@ def hash_video(
...
+        else:
+            ranges.append(f"{start}-{end}" if start != end else str(start))
+            start = idx
+            end = idx
+
+    ranges.append(f"{start}-{end}" if start != end else str(start))
+    return ranges
+
+
+def extract_mismatch_details(video_path: Path, csv_df: pd.DataFrame | None) -> dict:
```
sneakers-the-rat: re: comment in CLI on its usage, i think this function and `format_missing_frame_ranges` can be removed by just emitting the information needed in the exceptions themselves - they're just used for user feedback, so we just need enough information to create an error message (which the exception should just do)

tests/test_process/test_stitch.py
```diff
@@ -0,0 +1,374 @@
...
+
+def test_crop_video_hash(tmp_path):
+    """Cropped video content matches expected hash."""
+    out = crop_run(
+        str(STITCH_DATA_DIR / "video1.avi"),
+        output_path=str(tmp_path / "cropped.avi"),
+        trim_start=10,
+        trim_end=10,
+    )
+    assert hash_video(out) == EXPECTED_CROP_VIDEO_HASH
```
sneakers-the-rat: it's a *little* indirect, but i'll trust you validated the data here.

tests/test_process/test_stitch.py
```diff
@@ -0,0 +1,374 @@
...
+    _, is_tie = select_best_candidate([cand, cand])
+    assert is_tie is True
+
+
+def test_edge_scoring_selects_less_sharp():
+    """Sobel edge scoring picks the less sharp frame (higher score = less gradient)."""
+    uniform = np.ones((50, 50), dtype=np.uint8) * 128
+    edgy = np.zeros((50, 50), dtype=np.uint8)
+    edgy[:, 25:] = 255
+    assert score_edges(uniform) > score_edges(edgy)
```
sneakers-the-rat: nice

sneakers-the-rat (commented):
it's a chunky one, the broad strokes look good, but i think we can tighten this up a lot by thinking a lil bit about what we need here.

so we don't have a class or data model for "a mio dataset" really - a set of things like the acquisition config, and potentially multiple (video, metadata_csv) pairs. The existing video processing stuff works on a single (video, metadata_csv) pair, but this is the first time where having some structure around "a mio dataset" would really become useful as we're manipulating a bunch of different artifacts that we create. There is a start of what might become "mio dataset" in the `models.stitch` module, but those feel sort of ad-hoc, underutilized, and also in an odd location. 

my main note here then would be that if we just use that then i think we can simplify a lot of the verbose parts here - there is a ton of error and value checking all over the place, but if we did that once when creating "mio dataset" then we could both trim down on the boilerplate and also make things feel a bit more coherent.

I'm nearing the end of my working time tonight, but i'll get started working on that tomorrow.

t-sasatani:
Thanks for the review! I'll look through the comments, but I probably won't be able to fix anything in the coming week, so I would appreciate any refinements or fixes.

> my main note here then would be that if we just use that then i think we can simplify a lot of the verbose parts here - there is a ton of error and value checking all over the place

Yes, I totally agree it's quite redundant. Most of the checks were added when I was confused by `skvideo`. I would keep more verification than usual (maybe input and output), but I think we can remove many of them.

> if we did that once when creating "mio dataset" then we could both trim down on the boilerplate and also make things feel a bit more coherent.

Agreed on this as well. I wish there were a standard format for bundling videos and timestamps, and I wonder why it doesn't exist, since I guess it's quite a common need in scientific recording. I had similar issues while coding the placecell analysis package, and was wondering what the proper way to do this is. Did a structured directory with import/export methods for a self-defined extension, but it's becoming large and something very proprietary.
https://miniscope.github.io/placecell/_generated/placecell.dataset.html

mio/cli/process.py
```diff
@@ -2,10 +2,53 @@
...
 import click
 
+from mio.exceptions import VideoMetadataError
+from mio.io import VideoWriter
+from mio.logging import init_logger
 from mio.models.process import DenoiseConfig
-from mio.process.video import denoise_run
+from mio.process.stitch import RecordingData, RecordingDataBundle
+from mio.process.video import crop_run, denoise_run
+from mio.utils import (
```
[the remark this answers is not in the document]
t-sasatani: Makes sense, I end up using here as a Desktop-ish place, which obviously leads to a mess.

mio/cli/process.py
```diff
@@ -31,12 +74,432 @@ def process() -> None:
...
+    Processing steps (noisy area detection, frequency masking, minimum
+    projection subtraction) are configured via the YAML config file.
     """
+    csv_df = _validate_with_prompt(input)
+
     denoise_config_parsed = DenoiseConfig.from_any(denoise_config)
-    denoise_run(input, denoise_config_parsed)
+
+    if output_dir is not None:
+        denoise_config_parsed.output_dir = str(Path(output_dir).expanduser())
```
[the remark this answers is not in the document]
t-sasatani: Yes. I think I did this to add a suffix (the naming system may be a mess as well), but I assume that's still possible with Path objects.

mio/cli/process.py
t-sasatani: I think I had this earlier because I had misalignments by like one, which is an error, but was ok for me to move forward (which should have been a warning anyway). But surely looks awkward now, how weirdly user-facing it is. It does make more sense to just throw an exception.

mio/process/stitch.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @property
+    def metadata(self) -> pd.DataFrame:
+        """Get or load metadata CSV as DataFrame."""
+        if self._metadata is None:
+            self._metadata = pd.read_csv(self.csv_path)
+        return self._metadata
+
+
+class RecordingDataBundle:
```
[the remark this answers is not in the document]
t-sasatani: Yes that's better. I initially thought this will be data class kind of thing that gets plugged in to a processing method but yes this is a stitcher now.

mio/process/stitch.py
```diff
@@ -0,0 +1,278 @@
...
+    def stitch_recordings(self) -> None:
+        """Stitch recordings by iterating unique frame_nums and selecting the best frame."""
+        stitched_writes = 0
+        debug_writes = 0
+        frame_iter = tqdm(self.combined_frame_num, desc="Stitching frames")
+
+        for frame_num in frame_iter:
+            valid_pairs = self._collect_candidates(frame_num)
+            if not valid_pairs:
+                continue
```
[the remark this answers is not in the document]
t-sasatani: This should be for when a buffer exists, but a frame doesn't. This happens when the buffer doesn't meet the criteria and can't reconstruct a full frame.

mio/utils.py
t-sasatani: Yeah, I was confused while trying to show a summary of mismatches, but that can live in the exception, as you say.

mio/models/stitch.py
```diff
@@ -0,0 +1,74 @@
...
+import pandas as pd
+from pydantic import BaseModel
+from tqdm import tqdm
+
+from mio.logging import init_logger
+
+logger = init_logger(name="models.stitch")
+
+
+class FrameInfo(BaseModel):
```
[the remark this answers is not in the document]
sneakers-the-rat: ah ok, so could the names here be like "source_index" and "aligned_index" or something?

mio/models/stitch.py
sneakers-the-rat: ya ya i meant the pbar, i saw a bit more as i reviewed. i think we basically just want to move all this out of the methods themselves

mio/cli/process.py
```diff
@@ -31,12 +74,432 @@ def process() -> None:
...
+
+    Also trims and renumbers the companion CSV metadata to match.
+    """
+    csv_df = _validate_with_prompt(input)
+
+    input_path = Path(input)
+    output_arg = output if output is not None else DEFAULT_PROCESS_DIR
+    output_path = resolve_output_path(input_path, "_cropped", output_arg)
+
+    cropped_output = crop_run(
```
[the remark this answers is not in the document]
sneakers-the-rat: makes sense, we can just do this in the import :)

mio/process/video.py
```diff
@@ -587,3 +669,282 @@ def denoise_run(
...
+            "Successfully exported frame-timestamp CSV at %s (%d frames).",
+            output_csv_path,
+            len(frame_timestamps),
+        )
+    except Exception as e:
+        logger.error(f"Failed to write frame-timestamp CSV file {output_csv_path}: {e}")
+        raise
+
+
+def crop_run(
```
sneakers-the-rat: hmm wouldn't we want the outer union rather than the inner union? like if one video has frames 0-10, and the other has frames 2-12, we would want to keep 0-12 and just always take the first two frames from the first video and the last two frames from the second video?

t-sasatani: Yes, and the stitch should be using the outer union throughout. This is the "trim" method, so you might be looking at something else?

t-sasatani:
> Crop a video file by trimming frames from both ends.

I wasn't really thinking of the terminology difference here, but it feels pretty funny now that you mentioned it. Definitely something that an LLM would have handled better lol.

t-sasatani:
Sorry, just noticed I missed a lot of your comments for some reason, will look and reply within a few days.

mio/process/video.py
sneakers-the-rat: renamed, got it now

mio/data/config/process/denoise_example.yml
```diff
@@ -14,14 +14,14 @@ noise_patch:
     threshold: 20
   black_area_config:
     consecutive_threshold: 5
-    value_threshold: 16
+    value_threshold: 0
```
sneakers-the-rat, line 17: i see this change made it so we are adding another near-copy of this config in `tests/data/config/denoise_noise_detection_test.yml` - wondering what the purpose of this change/need for that addition is? i think without a description field on these it's hard for me to tell what the purpose of one config over another is.

mio/cli/process.py
```diff
@@ -31,12 +39,220 @@ def process() -> None:
...
+    help="Number of frames to remove from the start (default: 0).",
+)
+@click.option(
+    "-e",
+    "--trim-end",
+    type=int,
+    default=0,
+    help="Number of frames to remove from the end (default: 0).",
+)
+def workflow(
```
sneakers-the-rat, line 213: changes from the original here:
- removed all the error checking that happens in the middle: if there are errors we should raise them in the processing methods themselves, otherwise casting to `Recording` validates that the frame counts are identical between the video and the metadata csv
- didn't override the `output_dir` in the denoise config: we should use that if it's present, if it should output to another directory, the config should be changed.

otherwise should be the same. please try this out and see if it breaks your workflow, lmk what needs to change or what i have broken

sneakers-the-rat (approved):
ok returning to you @t-sasatani - managed to take about a thousand lines off while adding the dataset models, I didn't try and retrofit everything here to use them, but hopefully it's clear why they might be useful: we should avoid trying to juggle a million paths manually and should instead create a model of how all this stuff fits together. In the future this should also be some metadata file rather than being based on file naming conventions, but again trying to just set us up for future steps, and it ended up being simpler to add the dataset models now than having multiple different methods for validating and locating metadata files/etc.

I think the major observable difference here is in path manipulation, i am trying to go for a model where all the data for a given dataset stays in that directory, and for now a lot of that is flat with filenames to indicate what things are, but trying not to have multiple different directories where files from the same dataset are stored. once we get a better handle on how we want to organize files we can add back in more subdirectory structure within the context of a dataset.

give this a run and see if it does all you expect it to - the tests do pass, and they seem to test the major functionality, but i want to be sure i didn't break anything you are relying on (if there is anything that breaks when you run it, lmk and i can write more tests)

sneakers-the-rat:
if you're good with this, would you do the honors of writing the changelog and then feel free to merge :) otherwise lmk if changes requested

mio/cli/process.py
```diff
@@ -31,12 +74,432 @@ def process() -> None:
...
+    default=None,
+    help="Output path for debug CSV with selection metadata.",
+)
+@click.option(
+    "--fps",
+    type=int,
+    default=20,
+    help="Frames per second for output video.",
+)
+def stitch(
```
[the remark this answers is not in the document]
sneakers-the-rat: yeah i think we need more work on framerate generally, because to some degree the framerate in the video is made up and it is the timestamps that matter, but that's sort of a harder video encoding question. just keeping them the same between transcodes seems like a safe bet though

tests/test_cli_process.py
```diff
@@ -0,0 +1,179 @@
...
+
+import pytest
+from click.testing import CliRunner
+
+from mio.cli.process import process
+from mio.utils import hash_video
+
+STITCH_DATA_DIR = Path(__file__).parent / "data" / "stitch"
+
+EXPECTED_STITCHED_VIDEO_HASH = "69217a3079908094e11121d042354a7c1f55b6482ca1a51e1b250dfd1ed0eef9"
```
t-sasatani: Noticed the output file name was different, and this hash you changed to seems to be a hash for an empty output.

```
>>> import hashlib
>>> print(hashlib.new('blake2s').hexdigest())
69217a3079908094e11121d042354a7c1f55b6482ca1a51e1b250dfd1ed0eef9
```

t-sasatani:
I haven't read through all the changes yet, but I at least ran the script locally and made some fixes.
The stitching part is fairly decoupled from the other parts, so I think it's good to merge it. The following are more like notes. Might be worth just showing better error messages, but also not critical:

- This is an upstream issue, but there are some datasets where the metadata and frame count are off by one (don't know yet if this is a termination issue or anything else). I think I made a small prompt asking whether the code should move forward, but the current version just stopping is probably better.
- The code fails when the output filepath exists. My typical way to run this is to change some processing parameters and rerun the same command, expecting to overwrite, so I got confused for a moment. But I guess this is me overfitting to a suboptimal interface, and I'm not sure what a good interface would be.

sneakers-the-rat:
> there are some datasets where the metadata and frame count are off by one 

As in the metadata has an extra frame that is not in the video? Or what? Maybe we just trim that from the metadata? 

> The code fails when the output filepath exists.

What about a `--force` flag that enables overwriting?

tests/test_cli_process.py
sneakers-the-rat: Holy shit good catch lmao thank you

t-sasatani:
> the metadata has an extra frame that is not in the video? 
Yes, and it's tricky that I have no idea how to reproduce this. I think removing the extra metadata row before processing makes the most sense.

> What about a --force flag that enables overwriting?
Yes, that'll be perfect. Don't know why I didn't think of this.

sneakers-the-rat:
ok added those, will need to write tests for the --force flag and whatnot

sneakers-the-rat:
ok i'm gonna merge this puppy and please raise issues with anything that comes up!!!

PR #140, opened 2025-11-05
Device parameter bundle update

t-sasatani:
# Summary
- PR for batch uploading update commands to multiple devices

# Batch command file
A YAML file defines the uploaded commands.
```yaml
id: device-command-example
mio_model: mio.models.update.UpdateBatch
mio_version: 0.8.2.dev13+g338e847.d20251030
devices:
- device_id: 5
  updates:
    LED: 5
    GAIN: 4
    ROI_X: 20
    ROI_Y: 30
    SUBSAMPLE: 0
- device_id: 2
  updates:
    LED: 10
    GAIN: 1
    ROI_X: 60
    ROI_Y: 60
    SUBSAMPLE: 0
```

# CLI
```bash
# This will upload the above configuration via a connected IR remote control device.
mio update -b device-command-example
```
**This is more likely to use a local YAML file because it's user/experiment-dependent rather than device-dependent.**
```bash
mio update -b custom-setting.yaml
```

# Terminal output
```
[25-11-05T17:06:57] INFO     [mio.device_update] Updating device id 5 with LED=5                                                                              device_update.py:51
[25-11-05T17:06:58] INFO     [mio.device_update] Updating device id 5 with GAIN=4                                                                             device_update.py:51
[25-11-05T17:06:59] INFO     [mio.device_update] Updating device id 5 with ROI_X=20                                                                           device_update.py:51
[25-11-05T17:07:00] INFO     [mio.device_update] Updating device id 5 with ROI_Y=30                                                                           device_update.py:51
[25-11-05T17:07:01] INFO     [mio.device_update] Updating device id 5 with SUBSAMPLE=0                                                                        device_update.py:51
[25-11-05T17:07:02] INFO     [mio.device_update] Updating device id 2 with LED=10                                                                             device_update.py:51
[25-11-05T17:07:03] INFO     [mio.device_update] Updating device id 2 with GAIN=1                                                                             device_update.py:51
[25-11-05T17:07:04] INFO     [mio.device_update] Updating device id 2 with ROI_X=60                                                                           device_update.py:51
[25-11-05T17:07:05] INFO     [mio.device_update] Updating device id 2 with ROI_Y=60                                                                           device_update.py:51
[25-11-05T17:07:06] INFO     [mio.device_update] Updating device id 2 with SUBSAMPLE=0                                                                        device_update.py:51
```

----
📚 Documentation preview 📚: https://miniscope-io--140.org.readthedocs.build/en/140/

### DIFF
--- a/mio/device_update.py
+++ b/mio/device_update.py
@@ -36,7 +36,7 @@ def device_update(
     """
 
     if port:
-        logger.info(f"Using port {port}")
+        logger.debug(f"Using port {port}")
     else:
         ftdi_port_list = find_ftdi_device()
         if len(ftdi_port_list) == 0:
@@ -45,17 +45,17 @@ def device_update(
             raise ValueError("Multiple FTDI devices found. Please specify the port.")
         if len(ftdi_port_list) == 1:
             port = ftdi_port_list[0]
-            logger.info(f"Using port {port}")
+            logger.debug(f"Using port {port}")
 
     command = DevUpdateCommand(device_id=device_id, port=port, key=key, value=value)
-    logger.info(f"Updating {key} to {value} on port {port}")
+    logger.info(f"Updating device id {device_id} with {key}={value}")
 
     try:
         serial_port = serial.Serial(port=command.port, baudrate=2400, timeout=5, stopbits=2)
     except Exception as e:
         logger.exception(e)
         raise e
-    logger.info("Open serial port")
+    logger.debug("Open serial port")
 
     try:
         id_command = (command.device_id + UpdateCommandDefinitions.id_header) & 0xFF
@@ -88,7 +88,7 @@ def device_update(
 
     finally:
         serial_port.close()
-        logger.info("Closed serial port")
+        logger.debug("Closed serial port")
 
 
 def find_ftdi_device() -> list[str]:


sneakers-the-rat:
>  this downlink is very slow (like 1 second for sending two 8-bit words) due to demodulation limitations 

ah, ok, so we can't just send all the config parameters every time. good to know.

> So I think the ideal form would be like:

Yep! that's pretty much what i'm thinking too.

So we have the notion of a "device" which corresponds to some DAQ class and some config. 

That config consists of several different kinds of parameters
- **constant** parameters that are *always* true for a given device (its device name, stuff like its bit depth and other things that are fixed by the hardware/device implementation). these can't be changed by the user
- **firmware** parameters that are reflective of a given combination hardware/firmware/FPGA code - these can only be changed by the user by changing the firmware, but for a given firmware they are constant.
- **ephemeral** or **state** parameters (idk what to call this exactly, but these are the ones you are supposed to be able the change, the knobs and sliders and so on)
- **runtime** parameters that control how `mio` behaves in a way that's separable from the device behavior.

imo a device should only have one place where these are all configured, even if some of them are just references (so the config doesn't have to be huge and get out of date and stuff). The model we have so far of a declarative configuration i think is a good one, and i think doesn't really conflict with imperative updates/changes to the device, it's just a matter of how we present it as a ui: setting `{LED: 10, gain: 1, ...}` in your `device.yaml` file and running `mio update my-device-id` could be the same implementation as defining properties in an `UpdateBatch`, but it's presented as being part of the general device config. 

the thing we want to avoid is for that single place to be *misleading* or *wrong* - we don't want to tell someone they can set parameters if they can't, or if the device has a different value than the intended value.

so if we can't just write the params to the device on startup everytime because sending is slow, but receiving from the device is fast, the next best thing would be to have the device tell us what its current values are for these "ephemeral config" fields, and then we compare that to what we have in our config and only send the things that are different. we'd need the ability to do that anyway to know if the commands we sent were successful anyway. so could it work something like this?
- when starting to capture, mio sends some `getConfig` command to the device
- in its next header that it sends, it includes some key/value series with config names and values as an appendix to the usual header
- mio knows to parse these extra values when they're present, and makes them available via the `StreamDaq` class
- (then the rest of it, we send any commands whose config value is different from the device's current value)

that way we have correct information about the device state while we're recording, the config file 'just works' and serves as the source of truth, and also minimize the number of commands we have to send to the device?

>  -   Update via GUI sliders while looking at the video preview.
>  -   Output recording data, metadata, configs, and this user-mutable settings as a structured output object.
>  -   Make the GUI slider states importable from prior recordings.
>
> But if you're thinking of mio replacing the v4 Miniscope DAQ, I think we need to redesign this update stuff.

So yes how i would think about the GUI/TUI part of this would be

the device config would look something like this, mostly the same, just moving the params the user can't change somewhere else and getting them by reference, and then adding the config values from this PR as an example:

```yaml
id: my-device
mio_model: mio.models.stream.StreamDevConfig
mio_version: "v5.0.0"

# have the values currently in the device config defined elsewhere, 
# the device type and firmware config values are essentially constant to us
device_type: wireless-miniscope-or-whatever-its-name-is
firmware: "XEM7310-A75/USBInterface-8_33mhz-J2_2-3v3-IEEE.bit"

# this is the stuff we're intended to touch
runtime:
  serial_buffer_queue_size: 10
  frame_buffer_queue_size: 5
  image_buffer_queue_size: 5
  # ... etc.
  
# the config values we can change while running
state:
  led: 10
  gain: 1
  roi_x: 60
  roi_y: 60
  subsample: 0
```

so then w would have either one `set(key, value)` method or several `set_led(value)` methods that work something like this:

```python
def set_led(value: int) -> int:
    """Set the LED to some value, returning the value the device responded with"""
    
    # prevent other calls to change the config at the same time
    with self._config_lock:
        # send the update command to the device
        self.send_device_config("led", value)
    
        # wait until it returns the new value of `led` 
        # (confirming it was set, or telling us it couldn't set the value in case e.g. it was out of range)
        device_value = self.await_config_value("led")
    
        # flush the value to the yaml file
        self.config.state.led = device_value
        self.config.update_yaml()
        
        return device_value
```

and then from the GUI you'd do something like this

```python
led_slider.onChange(streamdaq.set_led)
```

(or whatever, depending on how the gui was written) that called that `set_*` method when some slider was changed. so the config can both be changed by the user before running in the `yaml` file and also changed during capture (e.g. by a gui). there are a few steps to get there, we're working on those over in the `noob` package and it's coming along pretty well, e.g. we need to make the `capture()` call non-blocking, but i think that's a decent target for behavior.

> Thus, I'm unsure if interactivity will make stuff look worse.

totally reasonable, worth testing, and we could set whatever rate limits or whatever we want either in the device class or in whatever gui we write.

t-sasatani:
Are there fundamental roadblocks for making `mio` handle v4?
It does make sense to polish the I/O software in general, but the wireless is still at the "we need to get images stably" stage and will be somewhat limited for a while. So it might not be the best platform for demonstrating the richer I/O features you're interested in. Tho I won't oppose keeping it tidy for wireless.

It may make sense to adapt v4 first as these design discussions won't be an overkill, and I'm happy to help. I wasn't in the discussion for updating Miniscope QT DAQ, so I am not sure if you guys already discussed this.

sneakers-the-rat:
> A dedicated YAML file will work better for this.

yes! that's what i'm proposing too, just the existing yaml config file. 

> wasting uplink bandwidth.

well hopefully it only happens once on startup, and most of the time it does nothing because the values haven't changed. 

> Are there fundamental roadblocks for making mio handle v4?

not really at all. we're working on https://github.com/miniscope/noob to generalize graph processing for these devices and analysis, so just doing the stuff i was going to do in those `pipeline` PRs awhile ago in a different package. working on the last multiprocesing bit before porting over here, and then we have an actual roadmap towards making a reasonable design instead of the sort of hodgepodge it is now.

t-sasatani:
Yes, an easy-to-use asynchronous pipeline runner sounds great and should solve a lot of stuff.

> well hopefully it only happens once on startup, and most of the time it does nothing because the values haven't changed.

I was imagining reserving a slot in the recording metadata, which would continuously eat bandwidth. But it could be better in the long term to include a preamble or a word that indicates the packet type, so that different types of packets (recording data, some spontaneous data, ack?) can be sent.

t-sasatani:
Maybe you can approve if it feels good for now?

PR #141, opened 2025-11-22
create configs from cli

sneakers-the-rat:
opening as draft because it's left 4 dead time, but @t-sasatani here's an initial draft of some cli commands for creating configs and opening them and whatnot.

so you should get like

```
mio config create --list
                            mio config models                            
┏━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ name                     ┃ module path                                ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ BufferHeaderFormat       │ mio.models.buffer.BufferHeaderFormat       │
│ DenoiseConfig            │ mio.models.process.DenoiseConfig           │
│ FreqencyMaskingConfig    │ mio.models.process.FreqencyMaskingConfig   │
│ SDBufferHeaderFormat     │ mio.models.sdcard.SDBufferHeaderFormat     │
│ SDLayout                 │ mio.models.sdcard.SDLayout                 │
│ StreamBufferHeaderFormat │ mio.models.stream.StreamBufferHeaderFormat │
│ StreamDevConfig          │ mio.models.stream.StreamDevConfig          │
│ UpdateBatch              │ mio.models.update.UpdateBatch              │
└──────────────────────────┴────────────────────────────────────────────┘
```

and then you can do like

```
mio config create UpdateBatch my-batch \
  -v 'devices=[{"device_id": 0, "updates": {"LED": 1}}]'

Wrote UpdateBatch config to /Users/jonny/Library/Application Support/mio/config/my-batch.yaml
id: my-batch
mio_model: mio.models.update.UpdateBatch
mio_version: 0.8.2.dev18+g3d8d838.d20251122
devices:
- device_id: 0
  port: null
  updates:
    LED: 1

```

and etc. but as you can see that's not all that elegant, all the models have been written with required parameters (since they should have required parameters), but we might need a little bit of a wrapper that emits an "empty version" of each of the models: e.g. for something like those update batches, since they're a list of another model, we would have to know "make one example empty model when the type is a list of another model type..' which is pretty cumbersome. so if we want to have easy cli config creation we'll need to change up the way we're modeling things (we do) or write some wrapping code for that.

what i also figure we'll want is making a copy of one config, e.g. one wants to make a derivative of the wireless scope they could do `mio config copy wireless-20px my-custom-wireless` so then we avoid needing to fill in a ton of defaults every time.

----
📚 Documentation preview 📚: https://miniscope-io--141.org.readthedocs.build/en/141/

### DIFF
--- a/mio/cli/config.py
+++ b/mio/cli/config.py
@@ -8,7 +8,6 @@ import platform
 import re
 import subprocess
 from pathlib import Path
-from typing import Optional
 
 import click
 import yaml
@@ -221,11 +220,11 @@ def _list(verbose: int) -> None:
 )
 @click.option("--list", "show_list", is_flag=True, help="List available models")
 def create(
-    model: Optional[str] = None,
-    config_id: Optional[ConfigID] = None,
+    model: str | None = None,
+    config_id: ConfigID | None = None,
     force: bool = False,
     value: tuple = (),
-    output: Optional[Path] = None,
+    output: Path | None = None,
     show_list: bool = False,
 ) -> None:
     """


sneakers-the-rat:
idk why the linter decided today was the day to format the tests but hey that's why we have linters and formatters lol. idk why those were excluded.

sneakers-the-rat:
alright should be good to review. added the `open` call, but not exactly sure how to test that

t-sasatani:
I think this is good as is to merge.

But I think a typical CLI usage pattern looks like this, so `copy` is pretty critical for usability.
```bash
mio config copy wireless-200px my-custom-wireless
mio config open my-custom-wireless
[open and save in the editor]
```
It's pretty complex to figure out the keys for the batch interface, and there are quite a few fields to fill in.

Another pattern that might be useful is something for locally sharing, like,
```bash
mio config import example-config.yml
```
but this could be an overkill for the current CLI-based thing.

t-sasatani (approved):

sneakers-the-rat:
> It's pretty complex to figure out the keys for the batch interface

ya agreed, we need to redesign the models to split up the different levels of config as we discussed before so that one would only need to set the essential things there. 

> But I think a typical CLI usage pattern looks like this, so copy is pretty critical for usability.

OK sure I'll do that when I get back. That should be like 15 lines.

> Another pattern that might be useful is something for locally sharing

Not sure what you mean? Just need to copy a file into your configs directory.

should be pretty much like

```
mio config user create ~/wherever
cp example-config.yml ~/wherever/config/
```

PR #142, opened 2025-11-26
Helpers for NTP server sync

t-sasatani:
This pull request adds a module that checks whether the system is synchronized with an optional NTP server. This doesn't perform the actual sync because it's OS-dependent and requires `sudo`. Any additions or suggestions are welcome. For instance, I have no idea how to test this.

- Users can define the NTP server address and the allowed time offset in runtime configurations for various CLI commands (currently integrated only with the `stream` command).
- The `prompt_ntp_sync` method measures the offset between the system clock and the specified NTP server. This method can be integrated with CLI commands.
- The NTP check can be bypassed through interactive prompts. This feature is needed because NTP servers may not be connected. 

# Example
Add something like below to the CLI command
```python
    if config.runtime.ntp_server is not None:
        prompt_ntp_sync(
            config.runtime.ntp_server, max_offset_seconds=config.runtime.ntp_max_offset_seconds
        )
```

If the specified NTP server exists and the time is synced, it proceeds:
```bash
[25-11-26T10:37:05] INFO     [mio.ntp] Checking time sync with NTP server: mio-ntp.local                       ntp.py:78
[25-11-26T10:37:22] INFO     [mio.ntp] Resolved mio-ntp.local to 192.168.10.2                                  ntp.py:33
                    INFO     [mio.ntp] Time is synchronized with NTP server mio-ntp.local (offset: 0.000s)     ntp.py:89
```

And if not, a query shows up:
```bash
[25-11-26T10:38:40] INFO     [mio.ntp] Checking time sync with NTP server: mio-ntp.local                       ntp.py:78
[25-11-26T10:38:57] WARNING  [mio.ntp] Could not resolve hostname mio-ntp.local to IP address.                 ntp.py:36
[25-11-26T10:38:59] WARNING  [mio.ntp] Could not query NTP server mio-ntp.local.                               ntp.py:82
System time may not be synchronized. Proceed anyway? [y/N]:
```

----
📚 Documentation preview 📚: https://miniscope-io--142.org.readthedocs.build/en/142/

### DIFF
--- a/docs/meta/changelog.md
+++ b/docs/meta/changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Upcoming
+- [`#142`](https://github.com/miniscope/mio/pull/142) - Add NTP time synchronization feature. NTP is an optional dependency - install with `pip install mio[ntp]`
+
 ## 0.8
 
 ### 0.8.1 - 2025-09-25 - Bugfix for frame reconstruction, CLI updates, more metadata for `stream`


pyproject.toml
```diff
@@ -25,6 +25,8 @@ dependencies = [
     'typing-extensions>=4.10.0; python_version<"3.13"',
     "scikit-video>=1.1.11",
     "scipy>=1.13.0",
+    "ntplib>=0.4.0",
+    "pillow>=11.0.0,<11.3.0",
```
sneakers-the-rat: Why is pillow added here? I don't see it used

t-sasatani: Good catch, don't need it in this PR. Got mixed up while working on this at the same time, trying a bunch of camera-detection stuff that works on Windows too (which I gave up for now).
https://github.com/miniscope/mio/pull/144

 I think I ended up not needing `pillow` for this, too, so I will remove it.

t-sasatani:
Yes, that's exactly the few lines I was talking about. Will do if that looks better.

PR #144, opened 2025-11-26
Feature: support webcams for behavior recording

t-sasatani:
# Updates
- `mio/behavior_cam.py`: Multiprocess behavior cam runner. Very similar to `stream_daq`.
- `mio/io.py`: Add `cv2` backend to videowriter
- `mio/cli/usbcam.py`: CLI for behavior camera. Mostly matches the `stream capture` pattern, but has additional interfaces for camera listing and selection.
- `mio/data/config/camera/*.yaml`: Example config for camera
- `mio/devices/mocks.py`: Added `USBCamMock`. Similar to `OKDevMock` but uses `.npz` for data as this is mainly generated by the test fixture.
- `mio/devices/usbcam.py`: Helper functions for the camera.
- `mio/models/usbcam.py`: Model for the webcam.
- `mio/utils.py`, `mio/stream_daq.py`: Move `exact_iter` to `utils.py` for reusing.

# Misc
This started from porting functionalities of the following repo:
https://github.com/MarcelMB/ELP_behavior_wide_lens_camera

----
📚 Documentation preview 📚: https://miniscope-io--144.org.readthedocs.build/en/144/

### DIFF
--- a/tests/test_behavior_cam.py
+++ b/tests/test_behavior_cam.py
@@ -110,17 +110,15 @@ def test_usbcam_mock_read_all_frames(set_usbcam_input, tmp_path):
 
     assert frames_read == 5
 
-def test_usbcam_mock_set_get():
+def test_usbcam_mock_set_get(tmp_path):
     """USBCamMock.set()/get() should store and retrieve properties."""
     from mio.devices.mocks import USBCamMock
 
     # Need DATA_FILE set; use a minimal npz
-    import tempfile
-
-    with tempfile.NamedTemporaryFile(suffix=".npz") as f:
-        np.savez(f.name, frames=np.zeros((1, 2, 2, 3), dtype=np.uint8), timestamps=np.array([0.0]))
-        USBCamMock.DATA_FILE = Path(f.name)
-        mock = USBCamMock()
+    npz_path = tmp_path / "mock_props.npz"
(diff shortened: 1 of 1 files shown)


t-sasatani:
Thanks, that sounds great to me. The quality fixes on my end for getting this into main will probably happen after finishing the upcoming experiments, so no rush at all.

sneakers-the-rat:
I think ultimately the interface we will want is one where all the different types of cameras have a shared interface, and the config determines what class/etc. Is used to implement that interface. 

So we would want to be able to do `mio stream capture --device wireless-miniscope-1` as well as `mio stream capture --device usb-webcam-1` rather than have a separate command for each type of device. As you say, there is more overlap than non-overlap in the code, maybe the only thing that's unique is the initial stage of pulling data. I think the exception might be the SD-card scopes, which I think there is only one of, because they're not streaming devices.

then stuff for specific device types would be like `mio device usbcam --list` and `mio device usbcam create` and whatnot.

BC eventually we will want to be able to do like `mio stream capture --device cam1 --device cam2` and so on to do simultaneous sync'd capture, and when making a GUI, having a common programming interface will be very very useful.

t-sasatani:
Makes sense. I think about the combining thing a bit later.

PR #155, opened 2026-03-16
Add noise-aware stitching, min_rows, concat, and calcium imaging config

MarcelMB:
## Summary

Adds several features for processing dual-DAQ wireless miniscope recordings, motivated by a 9-hour calcium imaging session with two DAQs.

### Noise-aware frame selection (`--selection-mode noise_aware`)
- During stitching, runs mio's existing `InvalidFrameDetector` (gradient + black_area) on each candidate frame
- If one DAQ's frame is noisy and the other is clean → pick the clean one
- If both are clean → pick the first (no expensive Sobel computation)
- If both are noisy → **skip the frame entirely** (no corrupted data in output)
- Produces terminal summary with per-DAQ noise percentages and both-broken stats
- Generates `noise_report.png` (timeline, run length distribution, drop density plots)
- Writes `both_broken.avi` for manual inspection of skipped frames

### `min_rows` parameter for `BlackAreaDetector`
- New `min_rows` field on `BlackAreaDetectorConfig` (default=1, preserves existing behavior)
- Requires N flagged rows before marking a frame as invalid
- Eliminates false positives from naturally dark regions in calcium imaging (recommended: `min_rows=10`)

### Vectorized `BlackAreaDetector`
- Replaced pixel-by-pixel Python for-loop with numpy cumsum sliding window
- ~100x speedup (enables real-time noise detection during stitching at ~170 fps)

### Timestamp-based frame matching (`--match-by timestamp`)
- Matches frames across DAQs by nearest `buffer_recv_unix_time` within configurable threshold (default 25ms)
- Handles DAQs starting/stopping at different times naturally
- Alternative to existing `frame_num` matching (which remains the default)

### `mio process concat` command
- Concatenates sequential recording segments from one DAQ into a single AVI + CSV
- Discovers all AVIs in a directory with natural numeric sort
- Fuzzy CSV matching handles mismatched filenames (e.g., `long-8-002.avi` → `long-8.csv`)
- Renumbers `reconstructed_frame_index` to be contiguous across segments

### `--max-frames` flag
- Limits frame processing during stitching for quick test runs

### `denoise_calcium_imaging` config
- New example config with recommended calcium imaging parameters
- `consecutive_threshold: 30`, `value_threshold: 0`, `min_rows: 10`

## New output files

These are **additional** outputs created by the new features (on top of mio's existing outputs):

| File | Created by | Description |
|------|-----------|-------------|
| `noise_report.png` | `stitch --selection-mode noise_aware` | Per-DAQ noise timeline, run-length distribution, and drop density plots |
| `both_broken.avi` | `stitch --selection-mode noise_aware` | All candidate frames from frame pairs where both DAQs were noisy (skipped from output) |
| `<dir>_combined.avi` | `concat` | Single AVI merging all recording segments from one DAQ directory |
| `<dir>_combined.csv` | `concat` | Merged metadata CSV with contiguous `reconstructed_frame_index` |

## Example usage

```bash
# Concatenate segments per DAQ
mio process concat -d /path/to/neural_DAQ1/
mio process concat -d /path/to/neural_DAQ2/

# Full workflow: noise-aware stitch → denoise
mio process workflow \
  -i DAQ1_combined.avi \
  -i DAQ2_combined.avi \
  --match-by timestamp \
  --selection-mode noise_aware \
  -c denoise_calcium_imaging

# Quick test (first 50k frames only)
mio process workflow \
  -i DAQ1_combined.avi \
  -i DAQ2_combined.avi \
  --match-by timestamp \
  --selection-mode noise_aware \
  -c denoise_calcium_imaging \
  --max-frames 50000
```

## Backward compatibility

- All new features are opt-in via CLI flags
- Default matching remains `frame_num`, default selection remains `metadata`
- `min_rows` defaults to 1 (existing behavior)
- Existing tests unaffected

## Test plan

- [ ] `pytest tests/test_process/test_frame_helper.py -v` — min_rows + vectorized detection
- [ ] `pytest tests/test_process/test_stitch.py -v` — timestamp matching + concat
- [ ] Manual test on dual-DAQ recording with `--selection-mode noise_aware`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MarcelMB:
To-DO safeguard:
- adding frum_number to unix_timestamp matching (marcel approach) can make sense

- concat also should look into unix_time, not just file order (get first unix timestamp and latest), so also a guard

-seperate two steps noise filter and stitch as discussed above

mio/cli/process.py
```diff
@@ -177,6 +178,117 @@ def crop(
...
+)
+def concat(
+    directory: str,
+    output: Optional[str],
+    fps: int,
+) -> None:
+    """
+    Concatenate sequential recording segments from one DAQ into a single video.
+
+    Discovers all .avi files in the given directory (that have companion .csv files),
```
sneakers-the-rat, line 212: so about this "all" part - we may have other videos in the directory that are not to be concatenated, how about passing an explicit list of files?

mio/cli/process.py
```diff
@@ -177,6 +178,117 @@ def crop(
...
+    def _natural_sort_key(path: Path) -> list:
+        """Sort by splitting name into text and numeric parts for natural ordering."""
+        return [
+            int(part) if part.isdigit() else part.lower()
+            for part in re.split(r"(\d+)", path.name)
+        ]
+
+    avi_files = sorted(dir_path.glob("*.avi"), key=_natural_sort_key)
+
+    def _find_companion_csv(avi_path: Path) -> Optional[Path]:
```
sneakers-the-rat, line 230: the changes in the root branch standardize this

mio/cli/process.py
```diff
@@ -177,6 +178,117 @@ def crop(
...
+        RecordingData(video_path=avi, csv_path=csv) for avi, csv in valid_pairs
+    ]
+
+    first_input_path = valid_pairs[0][0]
+    output_arg = output if output is not None else DEFAULT_PROCESS_DIR
+    combined_video_path = resolve_output_path(first_input_path, "_combined", output_arg)
+    combined_csv_path = combined_video_path.with_suffix(".csv")
+
+    click.echo(f"Concatenating {len(recordings)} segments...")
+    concat_recordings(
```
sneakers-the-rat, line 276: this would definitely be something that would work well with the new `Recording` models -

```python
combined = recording1.concat(recording2, recording3, ...)
```

mio/cli/process.py
```diff
@@ -212,12 +324,51 @@ def crop(
     default=20,
     help="Frames per second for output video.",
 )
+@click.option(
+    "--match-by",
+    type=click.Choice(["frame_num", "timestamp"]),
```
sneakers-the-rat, line 329: i've separated `align` out to a separate action, but we can accept like `--align-kwargs` here or something

mio/cli/process.py
```diff
@@ -212,12 +324,51 @@ def crop(
...
+    "--timestamp-threshold",
+    type=float,
+    default=25.0,
+    help="Max time difference in ms for timestamp matching (default: 25).",
+)
+@click.option(
+    "--max-frames",
+    type=int,
+    default=-1,
+    help="Maximum number of frames to process. -1 means all frames. "
```
sneakers-the-rat, line 344: i am sort of dubious about adding this to main, seems pretty limited use

mio/cli/process.py
```diff
@@ -212,12 +324,51 @@ def crop(
...
+    "Useful for quick test runs.",
+)
+@click.option(
+    "--selection-mode",
+    type=click.Choice(["metadata", "noise_aware"]),
+    default="metadata",
+    help="Frame selection strategy. 'metadata' uses buffer count + edge scoring (default). "
```
sneakers-the-rat, line 351: imo this doesn't need parameterization, we should always make the best choices we know how to make, and in this case we can combine all the scoring information.

I'm a little confused about this though, because judging from the `workflow` command, it seems like stitching comes downstream from denoising, so by the time we are doing stitching all the noisy frames should have been dropped from the input recordings. isn't that the case? or do we want to support stitching pre-denoising

mio/data/config/process/denoise_calcium_imaging.yml
sneakers-the-rat, line 1: what does this do/how is it different than the other configs? another new config was added in #133 and i am trying to keep track of what they are

mio/models/process.py
```diff
@@ -101,6 +101,12 @@ class BlackAreaDetectorConfig(BaseModel):
         default=0,
         description="Pixel intensity value below which a pixel is considered 'black'.",
     )
+    min_rows: int = Field(
+        default=1,
+        description="Minimum number of flagged rows required to mark the frame as invalid. "
+        "Default of 1 preserves original behavior. For calcium imaging, values around 10 "
+        "reduce false positives from naturally dark regions.",
```
sneakers-the-rat, line 108: nice, thank you for the clear docstring

mio/process/frame_helper.py
```diff
@@ -215,47 +215,35 @@ def _detect_black_pixels(
...
             Tuple[bool, np.ndarray]: A boolean indicating if the frame is corrupted and noise mask.
         """
-        height, width = current_frame.shape
-        noisy_mask = np.zeros_like(current_frame, dtype=np.uint8)
+        consecutive_threshold = self.config.consecutive_threshold
+        black_pixel_value_threshold = self.config.value_threshold
+
+        # Boolean mask of "black" pixels, then cumsum-based sliding window
+        black_mask = (current_frame <= black_pixel_value_threshold).astype(np.float32)
+        cs = np.cumsum(black_mask, axis=1)
```
sneakers-the-rat, line 231: oh nice, this is certainly faster, but it doesn't necessarily do the same thing - this wouldn't reset the counter back to zero after a pixel that's above threshold. is that important?

mio/process/frame_helper.py
```diff
@@ -215,47 +215,35 @@ def _detect_black_pixels(
...
-                    logger.debug(
-                        f"Frame noisy due to {consecutive_count} consecutive black pixels "
-                        f"in row {y}."
-                    )
-                    noisy_mask[y, :] = 1  # Mark row as noisy
-                    frame_is_noisy = True
-                    break  # No need to check further in this row
+        if current_frame.shape[1] >= consecutive_threshold:
+            # Sliding window: sum of `consecutive_threshold` consecutive pixels
+            run_sum = cs[:, consecutive_threshold:] - cs[:, :-consecutive_threshold]
```
sneakers-the-rat, line 235: i don't think i understand this, why would we be subsetting and subtracting like this? aren't we just testing if any run goes over the threshold?

mio/process/stitch.py
```diff
@@ -67,6 +71,30 @@ def select_best_candidate(candidates: list[CandidateFrame]) -> tuple[int, bool]:
...
+def select_best_candidate_noise_aware(
+    candidates: list[CandidateFrame],
+) -> tuple[int, bool] | None:
+    """
+    Pick the best candidate using noise detection results.
+
+    Returns (best_index, was_tie) or None if all candidates are noisy (skip frame).
+
+    Logic:
+    - If all candidates are noisy, return None (skip this frame)
```
sneakers-the-rat, line 83: i think this mixes concerns with denoising - denoising is the operation that removes noisy frames, stitching assumes that we've already done that, and we're now just stitching the best available frames from the videos we have.

mio/process/stitch.py
```diff
@@ -246,27 +310,364 @@ def _finalize(self) -> None:
...
             pd.concat(self._metadata_parts, ignore_index=True).to_csv(
                 self.combined_csv_path, index=False
             )
 
-    def stitch_recordings(self) -> None:
-        """Stitch recordings by iterating unique frame_nums and selecting the best frame."""
-        stitched_writes = 0
-        debug_writes = 0
-        frame_iter = tqdm(self.combined_frame_num, desc="Stitching frames")
+    def _build_timestamp_matches(
```
sneakers-the-rat, line 320: i'll port this over to the changed code in the root branch.

i think what we'll eventually want to have is preserving more of the metadata from a recording along with its output so we know automatically when to do this, opened an issue #165

mio/process/stitch.py
```diff
@@ -246,27 +310,364 @@ def _finalize(self) -> None:
...
+            grouped = df.groupby("reconstructed_frame_index")["buffer_recv_unix_time"].max()
+            frame_indices = grouped.index.values
+            timestamps = grouped.values
+            sort_order = np.argsort(timestamps)
+            per_rec_timestamps.append((frame_indices[sort_order], timestamps[sort_order]))
+
+        ref_indices, ref_timestamps = per_rec_timestamps[0]
+        matches: list[dict[int, int]] = []
+
+        for i, (ref_idx, ref_ts) in enumerate(zip(ref_indices, ref_timestamps)):
```
sneakers-the-rat, line 350: i'll take a closer look at this later, but i hope we can find a way to vectorize this

mio/process/stitch.py
```diff
@@ -246,27 +310,364 @@ def _finalize(self) -> None:
...
+                    num_buffers=num_buffers,
+                    sum_black_padding=sum_black,
+                    metadata_rows=rows,
+                    edge_score=0.0 if skip_edge_score else score_edges(frame),
+                    is_noisy=self._detect_noise(frame),
+                )
+            )
+        return candidates
+
+    def _print_noise_summary(
```
sneakers-the-rat, line 402: yeah with these parts i think we really are exceeding what 'stitching' should do - this all seems like stuff that should happen during denoising (and i'd also really love to move all plotting into a subpackage so we can make matplotlib an optional dep). We *definitely* need a mechanism for keeping track of all the files that are generated/can be generated for a given set of data, that's part of the goal with the `Dataset` change.

mio/process/stitch.py
```diff
@@ -246,27 +310,364 @@ def _finalize(self) -> None:
...
+        max_frames : int
+            Maximum number of frames to write. -1 means all frames.
+        """
+        stitched_writes = 0
+        debug_writes = 0
+        skipped_both_noisy = 0
+        use_noise_aware = self._noise_detector is not None
+        match_position = 0
+
+        def _select(candidates: list[CandidateFrame]) -> tuple[int, bool] | None:
```
sneakers-the-rat, line 530: nuuu just marge functionality into one `select_best_candidate` fn

mio/process/stitch.py
```diff
@@ -246,27 +310,364 @@ def _finalize(self) -> None:
...
+                    self._per_rec_noisy[i] += 1
+
+        def _handle_both_noisy(candidates: list[CandidateFrame], position: int) -> None:
+            """Write all candidate frames to both-broken AVI for manual review."""
+            self._both_noisy_indices.append(position)
+            if self._both_noisy_writer is not None:
+                for c in candidates:
+                    self._both_noisy_writer.write_frame(c.frame)
+
+        if matching_method == "timestamp":
```
sneakers-the-rat, line 550: yep, figured we would need this eventually, and making alignment a separate step is in part needing to avoid this: we shouldn't have two big if/else blocks for something that should precede this operation

mio/process/stitch.py
```diff
@@ -246,27 +310,364 @@ def _finalize(self) -> None:
...
+        Path for the combined output CSV.
+    fps : int
+        Frames per second for the output video.
+    """
+    video_writer = VideoWriter(path=output_video_path, fps=fps)
+    metadata_parts: list[pd.DataFrame] = []
+    rfi_offset = 0
+    total_frames = 0
+
+    for i, rec in enumerate(tqdm(recordings, desc="Concatenating segments")):
```
sneakers-the-rat, line 645: this function looks fine, except progress bars should always be opt-in

tests/test_process/test_stitch.py
```diff
@@ -357,6 +358,93 @@ def test_edge_scoring_selects_less_sharp():
...
+    assert indices == list(range(len(indices)))
+
+    # Second segment's rfi should start after first segment's max rfi
+    df1 = pd.read_csv(STITCH_DATA_DIR / "video1.csv")
+    max_rfi_1 = df1["reconstructed_frame_index"].max()
+    # Combined CSV should have indices beyond max_rfi_1
+    assert df["reconstructed_frame_index"].max() > max_rfi_1
+
+
+def test_stitch_timestamp_matching(tmp_path):
```
sneakers-the-rat, line 410: this is a pretty expensive way to test this and it isn't specific - we need to construct a set of timestamps that test the basic problems/things that a time-based alignment method needs to do: find the closest neighbor, avoid cumulative drift from different framerates, etc.

sneakers-the-rat (commented):
let's merge #133 first and then change the base of this to `main`. I'll handle restructuring this to match the changes i made there. a few questions and notes, but i get the need for this and can handle tidying up

sneakers-the-rat:
OK pushing what i have for the day. things i still need to do

- figure out where the new stitching logic differs from the prior stitching logic in the tests
- validate the new black pixel counting method
- incorporate the timestamp-based alignment
- final cleanup

the diff got extremely fucked up from the removal of the MSE detector, but basically i simplified the noise-aware part of the stitching into that `score_noise` function, and so now that just creates another csv table alongside the video before stitching.

still TODO is to figure out what to do with the denoise processor classes, because they don't use the `_noise.csv` premade measurements at all - need to figure out which intermediate outputs we want to keep and then clean that all up - #173

sneakers-the-rat:
ok! so checking out the changes to the stitched video, they are correct! this PR differs from `main`'s stitching decisions on two frames:

- 2542: frame index on video 1=29, video 2=25
- 2544: frame index on video 1=31, video 2=27

this PR chooses video 2 for both of those where `main` chooses video 1.

It's because video 1 has big black missing areas in those frames! It has all 8 buffers, and the `black_padding_px` are both zero, so it just defaulted to video 1.

the indexes below are 0-indexed, so they are off by one from the frame indexes in the screenshots (which are 1-indexed)

| frame_num | video 1 index | video 2 index | `video_1.avi` | `video_2.avi` |
| --- | --- |--- | --- | --- |
| 2542 | 29 | 25 | Screenshot 2026-04-23 at 4 55 13 PM |  Screenshot 2026-04-23 at 4 56 27 PM |
| 2544 | 31 | 27 |  Screenshot 2026-04-23 at 4 55 38 PM | Screenshot 2026-04-23 at 4 56 54 PM |

so hooray! good test failures!

sneakers-the-rat (approved):

PR #177, opened 2026-04-25
Move headers to `devices`

sneakers-the-rat:
opening as draft to run ci - will describe later, but basically continuing refactoring into separate devices, and making pandera tables for the various metadata that we produce

decided to split this up into 2 prs for history's sake - this is the first part of the process: moving all the header stuff to the devices subpackage

----
📚 Documentation preview 📚: https://miniscope-io--177.org.readthedocs.build/en/177/


mio/models/frames.py
```diff
@@ -26,10 +26,7 @@ class BaseFrame(BaseModel):
     Pydantic model to store an image
     """
 
-    frame: FrameType | None = Field(
-        None,
-        description="Frame data, if provided.",
-    )
```
sneakers-the-rat, line 32: while working on moving the sdcard stuff around, there were some circular import errors between here and logging, so i took a look at the logging and it was for handling when these didn't exist. this model is basically just a container for a frame, so the frame should not be optional. the case where there are no frames/videos should be handled elsewhere

PR #179, opened 2026-04-27
BER measurement

t-sasatani:
Forgot I didn't merge this. I do want to archive this, but it can also be after noob.
Just wanted to open this to see how different it is.

----
📚 Documentation preview 📚: https://miniscope-io--179.org.readthedocs.build/en/179/


sneakers-the-rat:
So I take it BER is bitwise error rate?

This can get added in, no prob. I'm working in standardizing the metadata output now and can pick this up in another PR too. So it seems like what we want is to have some switches that control whether some metadata is present or not, can figure out how to do that in a general way.

mio/devices/stream/device.py
```diff
@@ -301,6 +301,54 @@ def _parse_header(self, buffer: bytes) -> tuple[StreamBufferHeader, np.ndarray]:
 
         return header_data, payload
 
+    def prbs15_ber(
```
sneakers-the-rat, line 305: A little more explanation or links to references re: what this is would be nice

mio/devices/stream/device.py
```diff
@@ -517,6 +565,7 @@ def capture(
         show_video: bool | None = True,
         show_metadata: bool | None = False,
         freq_mask_config: FrequencyMaskingConfig | None = None,
+        ber: bool = False,
```
sneakers-the-rat: So if I'm reading this right, this switches us into basically a different mode than normal capturing. Do we expect any more different "modes"? If so then we can just make this a `mode` param and by default it's `"stream"` or you can set it to Ber?

t-sasatani:
Yes, it's a Bit Error Rate measurement feature with Pseudo-Random data.

I need this feature pretty quickly, and (hopefully) this will be a small PR. So I'll go ahead and brush up the implementation, add one or so regression tests with a measured binary, and integrate your comments (like those I see about documentation and mode make total sense and would be very easy).

sneakers-the-rat:
Basically I think if you bundle it up into a mode switch and then put all the Ber stuff inside the main capture method into like a single `_ber_mode()` method (as in from capture we just call `_ber_mode()`, not as in dump all the code into that method) then that would be good for me. I'll be splitting this all up soon anyway and having this split off will make that easier.

tests/test_devices/test_stream.py
```diff
@@ -318,6 +318,43 @@ def test_metadata_plotting(tmp_path, default_streamdaq):
     )
 
 
+def test_ber_measurement(tmp_path, set_okdev_input):
```
sneakers-the-rat, line 321: I think part of the reason we are having problem with stuff breaking is that most of our tests are these big end to end tests - it would be great to test these functions at a lower level with known, generated data rather than having a bunch of magic number tests like this. like i can't really tell what is going wrong in the CI runs because i don't know what these numbers *should be* / what it means when the numbers are wrong.

mio/devices/stream/device.py
```diff
@@ -301,6 +302,162 @@ def _parse_header(self, buffer: bytes) -> tuple[StreamBufferHeader, np.ndarray]:
 
         return header_data, payload
 
+    def prbs15_ber(
```
sneakers-the-rat, line 305: i'm gonna want to split all these out into separable functions, bundling them all up as instance methods makes them very hard to test/reuse (can do this later)

sneakers-the-rat:
renamed to "capture" just to have one fewer additional names running around. "the default mode for capture is capture" is a little redundant but it's at least hopefully self explanatory

sneakers-the-rat (approved):

PR #191, opened 2026-04-29
Test perf

### LINKED ISSUE #172: unslow the tests
the tests are very slow. this is because we have a bunch of end-to-end tests that are operating on large-is input data. we should slim down the tests so that they run on whatever minimal subset of data that is needed to ensure they are working

sneakers-the-rat:
Fix: #172 

- raise default stdout log level to ERROR in tests, which is hella expensive and unused
- add a `n_frame` param to capture to only capture a few frames when that's all we care about
- decrease the timeout before we kill capture processes

----
📚 Documentation preview 📚: https://miniscope-io--191.org.readthedocs.build/en/191/


sneakers-the-rat:
alright increased test perf mostly by disabling logs and not reading things unnecessarily.

like ~2m improvement.

the biggest perf barrier remaining is just the time it takes to a) spawn processes and b) that takes a long time because our import time takes eons because of pydantic.

so once we transition to noob we can have a switch for running the streamdaq synchronously for testing (and i'm also not sure how much of a perf benefit multiprocessing is here anyway) then we can bring that down further.

PR #192, opened 2026-04-30
Convert core streamdaq routines to noob

sneakers-the-rat:
Converting the core of the streamdaq to noob, opening as a draft, this might take a day or two since we need to also add an additional feature to noob to ensure all the queues get drained when we hit 'stop'.

probably going to do this in two stages, first the core frame construction stuff, and then handle the video/csv writer/plotting stuff in a second PR

----
📚 Documentation preview 📚: https://miniscope-io--192.org.readthedocs.build/en/192/


mio/devices/stream/nodes.py
```diff
@@ -33,6 +39,61 @@
...
+def iter_fpga(config: StreamDevConfig) -> Generator[A[bytes, Name("chunk")], None, None]:
+    """
+    Iterate a raw binary stream from the FPGA in chunks
+    (not necessarily split into buffers)
+    """
+    # set up fpga interfaces
+    dev = init_okdev(config.bitstream, config.read_length)
+
+    while True:
+        yield next(dev)
```
sneakers-the-rat, line 51: i'm pretty sure this can just be `yield from dev`

docs/conf.py
sneakers-the-rat, line 1: necessary because of this: https://github.com/miniscope/mio/actions/runs/35676160003/job/106583019253#step:5:121

where sphinx is trying to resolve the reference in a function signature annotation like `def something() -> type[T]:` and it just goes like "idk there are lots of things named type" despite that being unambiguously named in the module context.

mio/cli/stream.py
```diff
@@ -84,6 +84,9 @@ def _capture_options(fn: Callable) -> Callable:
         "- 'capture' (default) capture video/metadata;\n"
         "- 'ber' runs a PRBS bit-error-rate test and produces no video/metadata output.",
     )(fn)
+    fn = click.option(
+        "--ntp", is_flag=True, help="Synchronize system clock with NTP before capturing"
+    )(fn)
```
sneakers-the-rat, line 89: seemed like it should be opt-in

mio/data/tubes/stream/stream-fpga.yaml
```diff
@@ -0,0 +1,37 @@
...
+    depends:
+      - chunk: fpga.chunk
+
+  # binary i/o
+  write_binary:
+    type: mio.io.append_binary
+    depends:
+      - path: input.capture_binary
+      - data: fpga.chunk
+
```
sneakers-the-rat, line 33: should change this to the `enable` that depends on input, this was a hack before that existed

mio/data/tubes/stream/stream-frame.yaml
```diff
@@ -0,0 +1,79 @@
...
+input:
+  config:
+    type: mio.devices.stream.StreamDevConfig
+    scope: tube
+  header_csv:
+    type: pathlib.Path
+    scope: tube
+  video_path:
+    type: pathlib.Path
+    scope: tube
```
sneakers-the-rat, line 16: probably also want video kwargs

mio/data/tubes/stream/stream-frame.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+  write_csv:
+    type: mio.io.BufferedCSVWriter
+    enabled: input.header_csv
+    params:
+      file_path: input.header_csv
+      header: input.header_cls
+    depends:
+      - data: trim.header
+      - reconstructed_frame_index: combine.frame_idx
```
sneakers-the-rat, line 66: getting these aligned is the reason for the sorta wonky graph structure, we could also just make the counter a separate node, but this works for now. we only know the frame index after combining buffers into a frame, since we make the very very permissive assumption of not knowing the number of buffers in a frame, but that can likely change.

mio/devices/stream/device.py
```diff
@@ -119,26 +118,20 @@ def _ber_mode(
 
     def capture(
         self,
-        read_length: int | None = None,
```
sneakers-the-rat, line 122: this should not be a parameter at capture time, this should be part of device configuration

mio/devices/stream/device.py
```diff
@@ -119,26 +118,20 @@ def _ber_mode(
...
         video: Path | None = None,
         video_kwargs: dict | None = None,
         metadata: Path | None = None,
         binary: Path | None = None,
         show_video: bool | None = True,
         show_metadata: bool | None = False,
         freq_mask_config: FrequencyMaskingConfig | None = None,
-        mode: Literal["capture", "ber"] = "capture",
-        ber_output: Path | None = None,
```
sneakers-the-rat, line 131: we should split this out into a separate method `bitwise_error_rate()`, that was hard before because so much of the machinery was inside the capture method, but that's no longer the case.

mio/devices/stream/device.py
```diff
@@ -161,151 +154,27 @@ def capture(
...
-
-        except KeyboardInterrupt:
-            self.logger.exception(
-                "Quitting capture, processing remaining frames. Ctrl+C again to force quit"
-            )
-            self.terminate.set()
+                "video_path": video,
+            },
+        )
+        runner = SynchronousRunner(tube)
```
sneakers-the-rat, line 171: should swap this out with ZMQRunner by default probably, but this would also be nice to be able to configure

mio/devices/stream/device.py
```diff
@@ -337,15 +206,15 @@ def _handle_frame(
                     self.logger.debug("Saving header metadata")
                     try:
                         meta_row = header.model_dump()
-                        self._buffered_writer.append(meta_row)
+                        self._buffered_writer.process(meta_row)
```
sneakers-the-rat, line 209: ah wait this whole method can also just go i think

mio/devices/stream/nodes.py
```diff
@@ -33,6 +36,181 @@
...
+
+        if not self._seen_start_buffer:
+            if header_data.frame_buffer_count != 0:
+                return MetaSignal.NoEvent, MetaSignal.NoEvent
+            else:
+                self._seen_start_buffer = True
+
+        if counter is not None:
+            header_data.buffer_recv_index = next(counter)
+        init_logger("parse_header").debug("HEADER: %s", header_data)
```
sneakers-the-rat, line 133: oop

mio/devices/stream/nodes.py
```diff
@@ -33,6 +36,181 @@
...
+        header_data, buffer = self.header_cls.from_buffer(chunk, self.config)
+
+        if not self._seen_start_buffer:
+            if header_data.frame_buffer_count != 0:
+                return MetaSignal.NoEvent, MetaSignal.NoEvent
+            else:
+                self._seen_start_buffer = True
+
+        if counter is not None:
+            header_data.buffer_recv_index = next(counter)
```
sneakers-the-rat, line 132: example of receiving something in a dependency-injection-like way using assets: some asset that persists through the whole session can be given to us to use, we don't have to carry it, in case this is something that could happen across nodes.

mio/devices/stream/nodes.py
```diff
@@ -33,6 +36,181 @@
...
+        init_logger("parse_header").debug("HEADER: %s", header_data)
+        return header_data, buffer
+
+
+class CombineBuffers(Node):
+    """Collect buffers until we the header tells us that we're in a new frame"""
```
sneakers-the-rat, line 138: typo

mio/devices/stream/nodes.py
```diff
@@ -33,6 +36,181 @@
...
+                frame = np.concatenate(buffers, axis=0).reshape(
+                    (self.config.frame_width, self.config.frame_height)
+                )
+            except ValueError as e:
+                raise DeviceConfigurationError(
+                    f"Could not reshape frame, "
+                    f"expected ({self.config.frame_width}, {self.config.frame_height}), "
+                    f"{self.config.frame_width * self.config.frame_height}px, "
+                    f"but buffers were {sum(len(b) for b in buffers)} pixels"
+                ) from e
```
sneakers-the-rat, line 181: turned this into an error from a warning - might be the wrong call, but we should have already padded the buffers previously, so them being a mismatched shape is a config error - the config that controls the trimming/padding has become unlinked somehow from the config that is doing the reshaping here.

mio/io/binary.py
```diff
@@ -0,0 +1,12 @@
+"""Raw bytes i/o"""
+
+from pathlib import Path
+
+
+def append_binary(path: Path | None, data: bytes) -> None:
```
sneakers-the-rat, line 6: this is the thing that we should change to just being disabled rather than having this weird optional path thing

mio/io/video.py
```diff
@@ -20,13 +21,12 @@ class VideoWriter:
...
         "-f": "avi",
         "-pix_fmt": "gray",
-        "-vsync": "0",
     }
 
     def __init__(
         self,
         path: str | Path,
-        fps: int,
+        fps: int = 30,  # FIXME: should be required, just defaulting for testing
```
sneakers-the-rat, line 29: jonny fix before merging

mio/plots/headers.py
```diff
@@ -138,7 +145,10 @@ def __init__(
 
         # If a single string is provided, convert it to a list with one element
         if isinstance(header_keys, str):
-            header_keys = [header_keys]
+            header_cls = resolve_python_identifier(header_keys)
+            if not issubclass(header_cls, BufferHeader):
+                raise TypeError("Header class must be a buffer header class!")
+            header_keys = header_cls.csv_header_cols()
```
sneakers-the-rat, line 151: this is sort of weird, but it's in line with the notion of a 'dataset' - like we eventually want to get to a point where a device is capable of declaring its output structure, so we do waht this to be a bit abstract. we aren't writing a generic csv writer, this is a csv writer for mio, and we get to be a little weird with it

tests/test_logging.py
```diff
@@ -70,7 +70,8 @@ def test_nested_loggers(capsys, tmp_path):
     warnings.warn(f"FILES IN LOG DIR: {list(log_dir.glob('*'))}", stacklevel=2)
     warnings.warn(f"ROOT LOGGER HANDLERS: {root_logger.handlers}", stacklevel=2)
 
-    assert len(root_logger.handlers) == 2
+    # the 2 we expect, and 2 from pytest for log capturing
+    assert len(root_logger.handlers) == 4
```
sneakers-the-rat, line 74: this happened when i updated the pytest version - now they have two handlers that get added for log capture

pyproject.toml
```diff
@@ -63,7 +64,10 @@ documentation = "https://miniscope-io.readthedocs.io/"
 
 [project.optional-dependencies]
 ntp = ["ntplib>=0.4.0"]
-plot = ["matplotlib>=3.9.2"]
+plot = [
+    "matplotlib>=3.9.2",
+    "noob[gui,zmq]>=1000.1.0",
```
sneakers-the-rat, line 69: jonny bump version

MarcelMB:
image

Did a test run with the complete hardware in a weird configuration. and no errors. clean bitstream. I only had issues with the mio repo bitstream which is probably only because I am working on different firmwares. But mio side is fine.

PR #194, opened 2026-05-06
Add packet error rate evaluation (PER) and SIGINT handling to BER mode

t-sasatani:
It's pretty much what the title says. The changes should be encapsulated within the BER methods and decoupled from everything else. The line counts got inflated a bit because I also changed the log messages to f-strings after getting confused.

----
📚 Documentation preview 📚: https://miniscope-io--194.org.readthedocs.build/en/194/


sneakers-the-rat (approved):
looks gewd to me. the f strings are fine there, that logging message gets called only once per run rather than e.g. once per frame or something, so no prob

